PR #2, opened 2023-05-25
Feature/GitHub ci

Manuthor:


.github/workflows/python_and_docker.yml
```diff
@@ -0,0 +1,62 @@
...
+    runs-on: ubuntu-latest
+    container: docker:19.03.15
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Login to GitHub Packages
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
```
ThibsG, line 27: reuse `${{ env.REGISTRY }}`?

.github/workflows/python_build.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3
+          cache: pip # caching pip dependencies
+
+      - name: Prerequisites
+        run: |
+          pip install maturin[patchelf]==0.14.9
+          sudo apt-get install --no-install-recommends -qq libclang-dev gcc-mingw-w64-x86-64
```
ThibsG, line 97: not a big deal, but maybe `sudo` is not required here

.github/workflows/python_and_docker.yml
Manuthor: thanks

.github/workflows/python_and_docker.yml
```diff
@@ -0,0 +1,62 @@
+---
+name: Docker
+
+on:
+  workflow_call:
```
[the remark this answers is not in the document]
Manuthor: it is required to call this workflow (as a reusable workflow). But you are right, calling a reusable workflow without parameters is not optimal

.github/workflows/python_build.yml
Manuthor: I tried without sudo but it is unfortunately required :/

PR #16, opened 2023-07-09
chore: update cloudproof to 2.1.1 (includes cover_crypt 12.0 and crypto_core 9.0)

Manuthor:


crate/kmip/src/kmip/data_to_encrypt.rs
```diff
@@ -27,6 +27,7 @@ pub struct DataToEncrypt {
 
 impl DataToEncrypt {
     /// Serialize the data to encrypt to bytes
+    #[must_use]
     pub fn to_bytes(&self) -> Vec<u8> {
         let mut bytes = Vec::new();
```
tbrezot, line 32: Use `with_capacity()` instead.

crate/kmip/src/kmip/kmip_objects.rs
```diff
@@ -150,6 +150,7 @@ impl Object {
     }
 
     /// Return the `KeyWrappingData` of that object if any
+    #[must_use]
```
tbrezot, line 153: This function is not expensive, is it really necessary?

crate/pyo3/python/scripts/test_kms.py
```diff
@@ -5,226 +5,228 @@
...
+            self.priv_key_uid,
         ) = await self.client.create_cover_crypt_master_key_pair(self.policy.to_bytes())
 
     async def test_master_keys(self) -> None:
         # Query public key from KMS
-        pubkey = await self.client.get_object(self.pubkey_uid)
-        self.assertEqual(pubkey.object_type(), "PublicKey")
-        self.assertIsInstance(PublicKey.from_bytes(pubkey.key_block()), PublicKey)
+        pub_key = await self.client.get_object(self.pub_key_uid)
+        self.assertEqual(pub_key.object_type(), 'PublicKey')
```
tbrezot, line 48: I am not a fan of the single quoted strings.

crate/server/src/core/kms.rs
```diff
@@ -119,10 +121,8 @@ impl KMS {
             };
 
             // Generate a new key
-            let key: Key<32> = {
-                let mut rng = self.rng.lock().expect("failed locking the RNG");
-                Key::<32>::new(&mut *rng)
-            };
+            let mut rng = CsRng::from_entropy();
```
tbrezot, line 124: If there is a `rng` in `self` you should use it instead as instantiating a new RNG is really expensive.

crate/server/src/core/kms.rs
```diff
@@ -535,7 +535,7 @@ impl KMS {
             )))
         }
 
-        // check if owner is trying to grant themself
+        // check if owner is trying to grant them self
```
tbrezot, line 538: I am not sure about this correction.

crate/server/src/database/cached_sqlite_struct.rs
```diff
@@ -19,7 +19,7 @@ pub struct KMSSqliteCacheItem {
     /// The handler to the sqlite
     sqlite: Arc<Pool<Sqlite>>,
     /// They key of the sqlite
-    key: Key<32>,
+    key: [u8; 32],
```
tbrezot, line 22: Can you use a constant from somewhere instead of `32`?

crate/server/src/database/cached_sqlite_struct.rs
```diff
@@ -38,7 +38,7 @@ pub struct KMSSqliteCacheItem {
 impl fmt::Debug for KMSSqliteCacheItem {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("KMSSqliteCacheItem")
-            .field("group_ id", &self.key)
+            // .field("group_id", &self.key)
```
tbrezot, line 41: daed code

crate/server/src/database/cached_sqlite_struct.rs
```diff
@@ -273,7 +273,8 @@ impl KMSSqliteCache {
                 let freeable_cache_id = freeable_sqlites.push(id)?;
 
                 // Add it to the SqliteCache
-                let mut item = KMSSqliteCacheItem::new(pool, key.to_owned(), freeable_cache_id);
+                let key: [u8; 32] = key.as_bytes().try_into()?;
+                let mut item = KMSSqliteCacheItem::new(pool, key, freeable_cache_id);
```
tbrezot, line 277: I would try to pass the `SymmetricKey` to `KMSSqliteCacheItem::new()` instead. Otherwise you create yet an other non-zeroizing copy of the key in memory.

One could argue that the KMS is intended to run in a safe place, which would be true. So I guess this is pretty much up to you :shrug:

crate/utils/Cargo.toml
```diff
@@ -20,6 +19,7 @@ serde_json = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 thiserror = { workspace = true }
+tiny-keccak = { version = "2.0.2", features = ["shake"] } # TODO(E.C): should be fixed in cosmian_crypto_core
```
tbrezot, line 22: What is this todo? Why can't you use the `kmac!` from crypto_core?

crate/utils/src/crypto/cover_crypt/encryption.rs
```diff
@@ -135,7 +131,7 @@ impl EncryptionSystem for CoverCryptEncryption {
         );
         Ok(EncryptResponse {
             unique_identifier: self.public_key_uid.clone(),
-            data: Some(ciphertext),
+            data: Some(ciphertext.to_vec()),
```
tbrezot, line 134: We actually may want to have two serializable traits: `Serializable` and `SecureSerializable`. That way encrypted objects would be serialized directly to `Vec<u8>` rather than `Zeroizing<Vec<u8>>`, which would avoid needing this clone.

crate/utils/src/crypto/curve_25519/encryption_decryption.rs
```diff
@@ -1,49 +1,38 @@
...
+pub const X25519_PRIVATE_KEY_LENGTH: usize = 32;
 
 impl EciesEncryption {
     pub fn instantiate(public_key_uid: &str, public_key: &Object) -> Result<Self, CryptoError> {
-        let rng = CsRng::from_entropy();
-
-        let public_key_bytes = public_key.key_block()?.key_bytes()?;
-        let public_key = X25519PublicKey::try_from_bytes(&public_key_bytes)?;
+        let public_key_bytes: [u8; X25519_PUBLIC_KEY_LENGTH] =
+            public_key.key_block()?.key_bytes()?.as_slice().try_into()?;
```
tbrezot, line 30: I would be surprised that you cannot store the deserialized key in a key block. Do you know why this is not done? Is this to follow KMIP?

crate/utils/src/crypto/curve_25519/encryption_decryption.rs
```diff
@@ -1,49 +1,38 @@
...
+        let public_key = X25519PublicKey::try_from_bytes(public_key_bytes)?;
 
         trace!("Instantiated hybrid ECIES encipher for public key id: {public_key_uid}");
 
         Ok(Self {
-            rng: Arc::new(Mutex::new(rng)),
```
tbrezot, line 46: You are losing performances by moving the RNG instantiation to the `encrypt` function. Is there a good reason for that?

crate/utils/src/crypto/curve_25519/operation.rs
```diff
@@ -120,10 +116,11 @@ pub fn create_ec_key_pair<R>(
 where
     R: CryptoRngCore,
 {
-    let key_pair = X25519KeyPair::new(rng);
-    let private_key =
-        to_curve_25519_256_private_key(&key_pair.private_key().to_bytes(), public_key_uid);
-    let public_key =
-        to_curve_25519_256_public_key(&key_pair.public_key().to_bytes(), private_key_uid);
+    // generate a key pair
```
tbrezot, line 119:
```suggestion
```

crate/utils/src/crypto/symmetric/aes_256_gcm.rs
```diff
@@ -72,10 +65,15 @@ impl EncryptionSystem for AesGcmSystem {
         };
 
         // recover key
-        let key = self.symmetric_key_key_block.key_bytes()?;
+        let key: [u8; Aes256Gcm::KEY_LENGTH] = self
+            .symmetric_key_key_block
+            .key_bytes()?
+            .as_slice()
+            .try_into()?;
```
tbrezot, line 72: This could be shorten by implementing the correct error conversion:
```suggestion
        let key: [u8; Aes256Gcm::KEY_LENGTH] =
            self.symmetric_key_key_block.key_bytes()?.try_into()?;
```
I think I have seen this pattern before. If you choose to change it, apply it everywhere.

crate/utils/src/crypto/wrap/encrypt_decrypt.rs
```diff
@@ -52,16 +47,13 @@ where
                         q_string,
                     } => match recommended_curve {
                         RecommendedCurve::CURVE25519 => {
-                            let public_key = X25519PublicKey::try_from_bytes(q_string).context(
+                            let q: [u8; X25519_PUBLIC_KEY_LENGTH] =
+                                q_string.as_slice().try_into()?;
```
tbrezot, line 51: Same here:
```suggestion
                            let q: [u8; X25519_PUBLIC_KEY_LENGTH] = q_string.try_into()?;
```
Or 
```
                            let q = <[u8; X25519_PUBLIC_KEY_LENGTH]>::try_from(q_string)?;
```

crate/utils/src/crypto/wrap/encrypt_decrypt.rs
```diff
@@ -118,18 +110,15 @@ pub fn decrypt_bytes(unwrapping_key: &Object, ciphertext: &[u8]) -> Result<Vec<u
                             d,
                         } => match recommended_curve {
                             RecommendedCurve::CURVE25519 => {
-                                let private_key =
-                                    X25519PrivateKey::try_from_bytes(&d.to_bytes_be()).context(
+                                let private_key: [u8; X25519_PRIVATE_KEY_LENGTH] =
+                                    d.to_bytes_be().as_slice().try_into()?;
```
tbrezot, line 114: same

crate/utils/src/lib.rs
```diff
@@ -29,16 +29,19 @@ pub trait DecryptionSystem {
 pub struct KeyPair(pub (Object, Object));
 impl KeyPair {
     /// Create a new `KeyPair` from a private and public key
+    #[must_use]
     pub fn new(private_key: Object, public_key: Object) -> Self {
         Self((private_key, public_key))
     }
 
     /// Get the private key
+    #[must_use]
```
tbrezot, line 38: I may not be on the same page with you about the use of `must_use`. To me, inexpensive functions don't have to be tagged.

crate/utils/src/types.rs
```diff
@@ -67,25 +66,22 @@ impl Serialize for ExtraDatabaseParams {
...
-    fn deserialize<D>(deserializer: D) -> Result<ExtraDatabaseParams, D::Error>
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: serde::Deserializer<'de>,
     {
         let bytes = <Vec<u8>>::deserialize(deserializer)?;
         let group_id = u128::from_be_bytes(bytes[0..16].try_into().unwrap());
-        let key = SymKey::from_bytes(bytes[16..48].try_into().unwrap());
-        Ok(ExtraDatabaseParams { group_id, key })
+        let key = SymmetricKey::try_from_bytes(bytes[16..48].try_into().unwrap()).unwrap();
```
tbrezot, line 83: This implementation seems strange to me: won't these `unwrap()` cause panics if we try to deserialize the wrong data?

tbrezot: The same goes for slicing over `bytes`: how can we be sure the `Deserializer` holds a serialized vector of at least 48 bytes?

tbrezot (commented):
Great job, lot's of useful cleanings.

Have you checked that replacing all these `pub(crate)` by `pub` declarations did not add anything to the API?

crate/cli/src/actions/cover_crypt/mod.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
     error::CliError,
 };
 
-/// Manage CoverCrypt keys and policies. Rotate attributes. Encrypt and decrypt data.
+/// Manage `CoverCrypt` keys and policies. Rotate attributes. Encrypt and decrypt data.
```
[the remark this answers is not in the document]
Manuthor: right

crate/cli/src/tests/shared/destroy.rs
```diff
@@ -24,7 +24,7 @@ use crate::{
 pub fn destroy(cli_conf_path: &str, sub_command: &str, key_id: &str) -> Result<(), CliError> {
     let args: Vec<String> = vec!["keys", "destroy", key_id]
         .iter()
-        .map(|s| s.to_string())
+        .map(|s| (*s).to_string())
```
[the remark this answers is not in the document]
Manuthor: indeed

crate/server/src/core/kms.rs
Manuthor: good point

crate/server/src/core/kms.rs
Manuthor: revert

crate/utils/Cargo.toml
Manuthor: kmip_utils.rs uses the new macro `kdf256` but cannot build without tiny-keccak

crate/utils/src/crypto/curve_25519/encryption_decryption.rs
Manuthor: reverted

crate/kmip/src/kmip/kmip_objects.rs
Manuthor: I must say I strictly apply clippy pedantic warnings:
```
error: this method could have a `#[must_use]` attribute
   --> crate/kmip/src/kmip/kmip_objects.rs:153:5
    |
153 |     pub fn key_wrapping_data(&self) -> Option<&KeyWrappingData> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: add the attribute: `#[must_use] pub fn key_wrapping_data(&self) -> Option<&KeyWrappingData>`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate
    = note: `-D clippy::must-use-candidate` implied by `-D warnings`

```

tbrezot:
> > @@ -20,6 +19,7 @@ serde_json = { workspace = true }
>  strum = { workspace = true }
>  strum_macros = { workspace = true }
>  thiserror = { workspace = true }
> +tiny-keccak = { version = "2.0.2", features = ["shake"] } # TODO(E.C): should be fixed in cosmian_crypto_core
>
> kmip_utils.rs uses the new macro `kdf256` but cannot build without tiny-keccak
Ah... indeed... would this be a patch release in CryptoCore? If it is, then
maybe we could patch this first?

crate/utils/Cargo.toml
tbrezot: Ah... indeed... would this be a patch release in CryptoCore? If it is, then maybe we could patch this first?

Manuthor: yes a patch release is a good idea

crate/utils/src/crypto/curve_25519/operation.rs
Manuthor: done

crate/pyo3/python/scripts/test_kms.py
Manuthor: There is a rule in `.pre-commit-config.yaml` named `double-quote-string-fixer` that does the fix everywhere. Should I remove it?

tbrezot: This is mostly a question of taste I think. But I would argue that double quoted strings are the "natural" way to write them.

So I would remove it. But if you don't, then don't :)

crate/server/src/database/cached_sqlite_struct.rs
Manuthor: thx

crate/utils/src/types.rs
Manuthor: the file disappears from last "tagging" PR

crate/utils/src/lib.rs
Manuthor: I understand from clippy that is recommended to use it:
https://rust-lang.github.io/rust-clippy/master/index.html#/must_use_candidate:
```
The lint only checks the arguments for mutable types without looking if they are actually changed. On the other hand, it also ignores a broad range of potentially interesting side effects, because we cannot decide whether the programmer intends the function to be called for the side effect or the result. Expect many false positives. At least we don’t lint if the result type is unit or already #[must_use].
```

crate/utils/src/crypto/wrap/encrypt_decrypt.rs
Manuthor: done here https://github.com/Cosmian/kms/pull/16/commits/7ebe6f13a6a2589f44430db0be9dd43175ac237b

crate/utils/src/crypto/wrap/encrypt_decrypt.rs
Manuthor: done here https://github.com/Cosmian/kms/pull/16/commits/7ebe6f13a6a2589f44430db0be9dd43175ac237b

crate/utils/src/crypto/symmetric/aes_256_gcm.rs
Manuthor: done here https://github.com/Cosmian/kms/pull/16/commits/7ebe6f13a6a2589f44430db0be9dd43175ac237b

crate/server/src/database/cached_sqlite_struct.rs
Manuthor: done here https://github.com/Cosmian/kms/pull/16/commits/7ebe6f13a6a2589f44430db0be9dd43175ac237b

Manuthor:
> Great job, lot's of useful cleanings.
> 
> Have you checked that replacing all these `pub(crate)` by `pub` declarations did not add anything to the API?

I've checked it now.
You are right about the `pub(crate)` removal: IMHO there is bug in clippy since `pub(crate) use` pattern is considered as private (even inside the crate): so clippy "fixes" the lint by replacing `pub(crate)` by `pub` then exposing too largely the module/function/object. 

Example:
```
error: pub(crate) import inside private module
 --> crate/cli/src/actions/shared/utils/mod.rs:7:1
  |
7 | pub(crate) use destroy_utils::destroy;
  | ----------^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | help: consider using: `pub`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pub_crate
  = note: `-D clippy::redundant-pub-crate` implied by `-D warnings`
```
With:
```
pub(crate) mod utils;
```
Module `utils` is not private (since it is declared as pub(crate)) and exposing objects in children of utils should be allowed at the same level of `pub(crate)`. 


I reverted all those clippy-nursery suggestions.

crate/pyo3/python/scripts/test_kms.py
Manuthor: For the time being, I will let it as single quote. And probably a dedicated PR will be better

crate/utils/Cargo.toml
Manuthor: creating this crypto_core issue: https://github.com/Cosmian/crypto_core/issues/50

crate/utils/src/crypto/cover_crypt/encryption.rs
Manuthor: creating crypto_core issue: https://github.com/Cosmian/crypto_core/issues/51

crate/kmip/src/kmip/data_to_encrypt.rs
Manuthor: done in https://github.com/Cosmian/kms/pull/16/commits/2926f421f66d1b589c1e33262ab0dd1d85a375aa

crate/server/src/database/cached_sqlite_struct.rs
Manuthor: done in https://github.com/Cosmian/kms/pull/16/commits/2926f421f66d1b589c1e33262ab0dd1d85a375aa

crate/utils/src/crypto/curve_25519/encryption_decryption.rs
Manuthor: But the `public_key` still needs to be deserialized, how would you do without deserialize it?

tbrezot: I am not sure, you can keep it as is.

PR #25, opened 2023-08-21
Optional attributes object_type

weblaetitia:


crate/cli/src/actions/shared/locate.rs
```diff
@@ -63,7 +63,8 @@ impl LocateObjectsAction {
     /// Export a key from the KMS
     pub async fn run(&self, client_connector: &KmsRestClient) -> Result<(), CliError> {
         // the object type is ignored
-        let mut attributes = Attributes::new(ObjectType::SecretData);
+        let mut attributes = Attributes::default();
+        attributes.object_type = Some(ObjectType::SecretData);
```
ThibsG, line 67: this can be rewrite without `mut`, something like:
```suggestion
        let attributes = Attributes {
            object_type = Some(ObjectType::SecretData);
            ..Default::default()
```

crate/server/src/core/operations/get_attributes.rs
```diff
@@ -39,7 +39,8 @@ pub async fn get_attributes(
         }
         Some(attrs) => attrs,
     };
-    let mut res = Attributes::new(object_type);
+    let mut res = Attributes::default();
+    res.object_type = Some(object_type);
```
ThibsG, line 43: same here

crate/kmip/src/kmip/kmip_types.rs
```diff
@@ -830,7 +830,8 @@ pub struct Attributes {
     /// symmetric key, etc.) SHALL be set by the server when the object is
     /// created or registered and then SHALL NOT be changed or deleted before
     /// the object is destroyed.
-    pub object_type: ObjectType,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
Manuthor, line 833: I have trouble finding in KMIP the reference where they say it is optional, is it from here https://docs.oasis-open.org/kmip/kmip-spec/v2.0/os/kmip-spec-v2.0-os.html#_Toc6497496?

crate/server/src/core/operations/get_attributes.rs
```diff
@@ -39,7 +39,10 @@ pub async fn get_attributes(
         }
         Some(attrs) => attrs,
     };
-    let mut res = Attributes::new(object_type);
+    let mut res = Attributes {
```
Manuthor, line 42: can't we keep the actual call?
```let mut res = Attributes::new(object_type);```

File: crate/kmip/src/kmip/kmip_types.rs [hunk 5]
```diff
@@ -849,22 +850,6 @@ pub struct Attributes {
 }
 
 impl Attributes {
-    #[must_use]
```
Manuthor, line 852: we could have keep this constructor. Why do you want to remove it?

crate/kmip/src/kmip/kmip_types.rs
Manuthor: if you find this ref, we could add it next to the `object_type`

ThibaudDauce: I didn't find it either but for `Locate` it is specified that `Attributes` could be empty
image

crate/kmip/src/kmip/kmip_types.rs
ThibaudDauce: Because it's an empty `new()` defaulting to all `None` that can be auto-generated by deriving the `Default` trait

crate/kmip/src/kmip/kmip_types.rs
Manuthor: ok nevermind

PR #32, opened 2023-08-28
Some improvements for the PKI branch

ThibsG:


crate/server/src/core/certificate/create_subca_certificate.rs
```diff
@@ -202,94 +216,60 @@ async fn locate_or_create_ca_signing_key(
...
-    let signer = match signing_keys_uids {
-        Some(uids) => match uids.len() {
-            0 => match profile {
-                Profile::Root => create_root_ca(ca, profile, tags, kms, owner, params).await?,
-                _ => {
-                    return Err(KmsError::ConversionError(format!(
-                        "Internal error: Profile root expected here for this CA: {ca}",
-                    )))
+    let Some(signing_keys_uids) = signing_keys_uids else {
+        return create_root_ca(ca, tags, kms, owner, params).await
```
ThibsG, line 220: @Manuthor maybe we could check for profile here. If the asked CA profile is not root, we must return an error and not generate a root CA, what do you think?

Manuthor: I agree, it misses a verification here

PR #37, opened 2023-09-12
Bulk encryption/decryption for Covercrypt

ThibsG:


crate/cli/src/actions/cover_crypt/encrypt.rs
```diff
@@ -82,18 +102,151 @@ impl EncryptAction {
...
-            .write_all(&data)
-            .with_context(|| "failed to write the encrypted file")?;
-
-        println!("The encrypted file is available at {:?}", &output_file);
+        if let Some(CryptographicAlgorithm::CoverCryptBulk) = cryptographic_algorithm {
+            self.write_bulk_encrypted_data(&data)?;
+        } else {
+            // Write the encrypted file
+            let output_file = if let Some(output_file) = &self.output_file {
+                output_file.clone()
```
tbrezot, line 110: You should be able to avoid cloning here since `File::create()` does not need it.

crate/cli/src/actions/cover_crypt/encrypt.rs
```diff
@@ -82,18 +102,151 @@ impl EncryptAction {
...
+            })?
+            .serialize()
+            .map_err(|_| {
+                CliError::Conversion(
+                    "Unable to serialize encrypted header structure to bytes".to_string(),
+                )
+            })?;
+        let encrypted_block = de.finalize();
+
+        let mut data_slice: &[u8] = encrypted_block.as_ref();
```
tbrezot, line 197: Why to create this variable?

crate/cli/src/actions/symmetric/encrypt.rs
```diff
@@ -69,8 +69,9 @@ impl EncryptAction {
             plaintext,
             None,
             self.authentication_data
-                .clone()
+                .as_deref()
                 .map(|s| s.as_bytes().to_vec()),
```
tbrezot, line 73: Is it KMIP that enforce storing an `Option<String>` in the `EncryptAction`? Because having to clone this is sad. I don't think we need to store it as `String` if KMIP does not need to.

tbrezot (commented):
Great job :)

- you should be able to avoid adding a dependency to `leb128` by using CryptoCore's serialization module;
- are we targeting 64-bit architectures only? because if we aren't, `u64 as usize` casts can panic when deserializing the wrong data.

crate/cli/src/main.rs
```diff
@@ -46,6 +46,13 @@ async fn main() {
 }
 
 async fn main_() -> Result<(), CliError> {
+    // Uncomment the following if you need debug logs
+    // from the CLI (useful for tests)
+    if option_env!("RUST_LOG").is_none() {
+        std::env::set_var("RUST_LOG", "debug");
+    }
+    env_logger::init();
```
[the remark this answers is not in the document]
ThibsG: yes, but I'll comment it, as it is interesting only when debugging

crate/cli/src/actions/cover_crypt/decrypt.rs
```diff
@@ -79,18 +93,131 @@ impl DecryptAction {
...
+            let mut f = File::open(input_file)
+                .with_context(|| format!("Can't read the file to decrypt ({input_file:?})"))?;
+            let mut content = Vec::new();
+            f.read_to_end(&mut content)
+                .with_context(|| "Fail to read the file to decrypt")?;
+
+            leb128::write::unsigned(&mut data, content.len() as u64).map_err(|_| {
+                CliError::Conversion("Cannot write the size of the chunk to encrypt".to_string())
+            })?;
+            data.append(&mut content);
```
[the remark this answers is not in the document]
ThibsG: the `Serializable` from CryptoCore is based on `Zerozing` which is not able to `append`, so there is an extra copy if we use it

crate/cli/src/actions/cover_crypt/decrypt.rs
```diff
@@ -79,18 +93,131 @@ impl DecryptAction {
...
+            let chunk_size = leb128::read::unsigned(&mut data_slice)
+                .map_err(|_| CliError::Conversion("Cannot read the chunk size".to_string()))?
+                as usize;
+
+            #[allow(clippy::needless_borrow)]
+            let chunk_data = (&mut data_slice).take(..chunk_size).ok_or_else(|| {
+                CliError::Conversion(
+                    "Unable to get a valid slice from decrypted response buffer".to_string(),
+                )
+            })?;
```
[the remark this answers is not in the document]
ThibsG: No copy with `::core::slice::take`

crate/cli/src/actions/cover_crypt/encrypt.rs
ThibsG: the prototype of `::core::slice::take` require the use of a dedicated `mut _: &[u8]` variable

crate/cli/src/actions/cover_crypt/encrypt.rs
ThibsG: yes but the fallback of this `if` cannot be a reference, hence I'm forced to clone here.

tbrezot: True. You could have avoided it by duplicating the code outside your if condition... but anyway this is a path, so cloning shouldn't be expensive. I didn't see it, I thought this was the actual file bytes.

You could have used something like `self.output.cloned().unwrap_or(|| self.intputs.first()...)` to avoid the if. But this is only style.

crate/cli/src/actions/cover_crypt/decrypt.rs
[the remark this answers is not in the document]
tbrezot: You can use the Serializer to manage everything. Thus said, it is true that the use of `Zeroizing` in the serializer was a mistake (mine). I am planning to add two serializers: `Serializer` (the plain old serializer), and `SecureSerializer` (which uses `Zeroizing` as output). Or maybe we should decide that serializing is already unsafe and that no more care should be taken to protect the key.

crate/cli/src/actions/cover_crypt/encrypt.rs
ThibsG: Ok in fact, I reworked this part in the same time you were reviewing, now it is the case:
https://github.com/Cosmian/kms/pull/37/files#diff-1254d8ddead7be701fb7158a92798d0b6b5f9ec46603fb9db99f5b502b4e4ba4R146

tbrezot (approved):

PR #39, opened 2023-09-20
:sparkles: add ratls cert in bootstrap server

heavenboy8:
Use RATLS cert in the bootstrap sever + use it (after verify) when querying the bootstrap server


Cargo.toml
```diff
@@ -37,6 +37,8 @@ leb128 = "0.2"
 native-tls = "0.2"
 num-bigint = { version = "0.4", features = ["rand", "serde"] }
 openssl = { version = "0.10", features = ["vendored"] }
+# ratls = { path = "/home/seb/dev/microservice_encryption/tee-tools/ratls" }
```
Manuthor: I let a small comment to not forget to clean this when ready

crate/cli/src/config.rs
```diff
@@ -53,6 +56,55 @@ fn not(b: &bool) -> bool {
...
+            self.signing_key,
+            self.sev_measurement,
+            guess_tee(),
+        ) {
+            (Some(e), Some(s), None, Ok(TeeType::Sgx)) => Ok(TeeMeasurement::Sgx {
+                mr_signer: compute_mr_signer(&fs::read_to_string(s)?)
+                    .map_err(|e| CliError::Default(format!("MR signer computation failed: {e:?}")))?
+                    .as_slice()
+                    .try_into()
+                    .map_err(|_| CliError::Default("Bad MR signer bytes-size".to_string()))?,
```
Manuthor, line 81: should we keep the error details?
```suggestion
                    .map_err(|e| CliError::Default(format!("Bad MR signer bytes-size: error: {e:?}"))?,
```

if we don't, maybe add a From conversion in the CLIError struct.

crate/client/src/bootstrap_rest_client.rs
```diff
@@ -209,9 +213,33 @@ impl BootstrapRestClient {
...
         };
 
+        // Get and verify the ratls certitificate in order to use it as the only valid root CA
+        let bootstrap_server_url = Url::parse(bootstrap_server_url)?;
+
+        let ratls_cert = get_server_certificate(
+            bootstrap_server_url
+                .host_str()
+                .ok_or(RestClientError::Default(
+                    "Missing 'hostname' in boostrap server url".to_string(),
```
Manuthor, line 223:
```suggestion
                    "Missing 'hostname' in bootstrap server url".to_string(),
```

crate/client/src/bootstrap_rest_client.rs
```diff
@@ -209,9 +213,33 @@ impl BootstrapRestClient {
             None => builder,
         };
 
+        // Get and verify the ratls certitificate in order to use it as the only valid root CA
```
Manuthor, line 216:
```suggestion
        // Get and verify the ratls certificate in order to use it as the only valid root CA
```

crate/client/src/bootstrap_rest_client.rs
```diff
@@ -209,9 +213,33 @@ impl BootstrapRestClient {
...
+
+        let ratls_cert = get_server_certificate(
+            bootstrap_server_url
+                .host_str()
+                .ok_or(RestClientError::Default(
+                    "Missing 'hostname' in boostrap server url".to_string(),
+                ))?,
+            bootstrap_server_url.port().unwrap_or(443) as u32,
+        )
+        .map_err(|e| RestClientError::RatlsError(e.to_string()))?;
```
Manuthor, line 227: either give some details of this error or simply add impl for `RestClientError`
```
impl From<ratls::error::Error> for RestClientError {
    fn from(e: ratls::error::Error) -> Self {
        Self::RatlsError(e.to_string())
    }
}
```

crate/client/src/bootstrap_rest_client.rs
```diff
@@ -209,9 +213,33 @@ impl BootstrapRestClient {
...
+
+        let ratls_cert = X509::from_der(&ratls_cert)
+            .map_err(|e| RestClientError::RatlsError(e.to_string()))?
```
Manuthor, line 230: same here

crate/server/src/bootstrap_server/certificate.rs
```diff
@@ -5,11 +5,37 @@ use openssl::{
...
 
+pub(crate) fn generate_ratls_pkcs12(
+    subject: &str,
+    expiration_days: u64,
+    pkcs12_password: &str,
+) -> KResult<Pkcs12> {
+    let (private_key, cert) = generate_ratls_cert(subject, vec![], expiration_days, None, true)
+        .map_err(|e| error::KmsError::RatlsError(e.to_string()))?;
+
+    std::fs::write("/tmp/cert.pem", &cert)?;
```
Manuthor, line 22: I don't know if we need to write this cert.pem.

If so, should we use something like:

```
    let tmp_dir = TempDir::new()?;
    let tmp_path = tmp_dir.path();
```

crate/server/src/config/command_line/bootstrap_server_config.rs
```diff
@@ -10,14 +10,22 @@ pub struct BootstrapServerConfig {
...
-    /// The name that will be the CN
-    /// in the bootstrap server self-signed certificate
+    /// Subject as an RFC 4514 string for the RA-TLS certificate
+    /// in the bootstrap server
     #[clap(
         long,
-        env("KMS_BOOTSTRAP_SERVER_COMMON_NAME"),
-        default_value("cosmian.kms")
+        env("KMS_BOOTSTRAP_SERVER_SUBJECT"),
+        default_value("CN=cosmian.kms,O=Cosmian Tech,C=FR,L=Paris,ST=Ile-de-France")
```
Manuthor, line 18: for my knowledge and just to make sure, this CN is only used temporary for the bootstrap process?

crate/cli/src/config.rs
heavenboy8: Do you prefer a From conversion ? I always hesitate to do that for conversion error because I could be use elsewhere (transparently) and we need more contexts most of the time. That's why I use a `map_err` on such error. What do you think about it ?

crate/server/src/config/command_line/bootstrap_server_config.rs
heavenboy8: Indeed

crate/cli/src/config.rs
Manuthor: Probably a From conversion is enough here.

PR #44, opened 2023-09-28
(Re) Enable bootstrap server for non "enclaves" servers

bgrieder:
A recent MR limits the use of the bootstrap server to "enclaves" and is doing an OS-dependent detection.
The bootstrap server may help supply configuration via API rather than on the command line, including on Linux.
Also, the PR was breaking tests for developers on MacOS.

This PR fixes these 3 issues. 

The mechanism can be further improved, and the following note was added in `bootstrap_server/certificate/mod.rs`

```
//! This module contains the code to generate a self-signed certificate for the server.
//!
//! On Linux: the code will attempt to generate a self-signed RA-TLS certificate.
//! If that fails and `ensure_ra_tls` is set the server will fail starting
//! otherwise it will generate a standard self-signed TLS certificate.
//!
//! On other targets, the code will generate a standard self-signed TLS certificate.
//! If `ensure_ra_tls` is set, the server will fail starting
//! with an error message explaining that RA-TLS is not available for this platform.
//!
//! The choice of making the detection OS-dependent is because RA-TLS is only supported on Linux via the ratls crate.
//! Intel and AMD stopped shipping drivers for other targets.

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ All notable changes to this project will be documented in this file.
 ### Bug Fixes
 
 - Improved database data structures using Maps and Sets instead of Vectors where uniqueness is required 
-
+- Enable bootstrap server for non "enclaves" servers
 
 ## [4.6.0] - 2023-09-01
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ env_logger = "0.10"
 hex = "0.4"
 http = "0.2"
 josekit = "0.8.3"
+leb128 = "0.2"
 native-tls = "0.2"
 num-bigint = { version = "0.4", features = ["rand", "serde"] }
 openssl = { version = "0.10", features = ["vendored"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,6 @@ env_logger = "0.10"
 hex = "0.4"
 http = "0.2"
 josekit = "0.8.3"
-leb128 = "0.2"
 native-tls = "0.2"
 num-bigint = { version = "0.4", features = ["rand", "serde"] }
 openssl = { version = "0.10", features = ["vendored"] }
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -49,6 +49,7 @@ env_logger = "0.10"
 hex = "0.4"
 http = "0.2"
 josekit = "0.8.3"
+leb128 = "0.2"
```
ThibsG, line 52:
```suggestion
```
also, need to re-run `cargo check` for `Cargo.lock` to be updated

crate/pyo3/Cargo.toml
```diff
@@ -14,7 +14,13 @@ cloudproof = { workspace = true }
 cosmian_kmip = { path = "../kmip" }
 cosmian_kms_client = { path = "../client" }
 cosmian_kms_utils = { path = "../utils" }
-pyo3 = { version = "0.18", features = ["extension-module", "abi3", "abi3-py37", "generate-import-lib"] }
+leb128 = { workspace = true }
```
ThibsG, line 17:
```suggestion
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [4.6.1] - 2023-09-28
```
ThibsG, line 5: let's keep this changelog diff 👍🏻 but no need to upgrade the version in features PR (releases and version bump must come from main), maybe more stuff is coming into develop :wink:

heavenboy8 (approved):

Cargo.toml
bgrieder: Done

crate/pyo3/Cargo.toml
bgrieder: Done

CHANGELOG.md
bgrieder: No problem with that but discipline must be enforced: changes that these PR further change went into main without ever being recorded in the changelog (and that was an API change)

PR #48, opened 2023-09-29
Improve error handling for KMIP data serialization

### LINKED ISSUE #46: Move serialization out of KMIP crate
In the KMIP create, the presence of the file `data_to_encrypt.rs` which is basically only serialization could be moved elsewhere.

Then the function could bubble-up appropriate errors instead of having an error variant about serialization problem in KMIP error enum.

ThibsG:
- Move the `DataToEncrypt` struct outside `kmip` crate to `utils` crate
- Use `Result` instead of unwrapping + rework calling sites

fixes #46


Manuthor (approved):
great move!

PR #50, opened 2023-10-02
feat: add certificate rsa support

Manuthor:


crate/utils/src/crypto/hybrid_encryption_system.rs
```diff
@@ -205,80 +234,71 @@ impl EncryptionSystem for EciesEncryption {
...
+            "Instantiated hybrid decryption system for user decryption key id: {private_key_uid}"
+        );
 
         Ok(Self {
             private_key_uid: private_key_uid.into(),
-            private_key_bytes: private_key.key_block()?.key_bytes()?,
-            recommended_curve,
+            private_key: private_key.clone(),
+            // recommended_curve,
         })
```
ThibsG, line 256: Looking at these changes, it doesn't seem to be interesting to have this `instantiate` anymore, what do you think?
One could just init `HybridDecryptionSystem` just by filling the fields

Manuthor: you are right, thank you

ThibsG (approved):
Thanks!

PR #63, opened 2023-10-10
Feat/edit policy

ackRow:


crate/cli/src/actions/cover_crypt/policy.rs
```diff
@@ -93,19 +91,39 @@ impl TryFrom<&Policy> for PolicySpecifications {
...
+                                .attributes
+                                .get(name)
+                                .map(|attr| match attr.encryption_hint {
+                                    EncryptionHint::Hybridized => "::+",
+                                    EncryptionHint::Classic => "",
+                                })
+                                .unwrap_or("")
+                    })
+                    .collect()
+            } else {
```
tbrezot, line 112: This is needed only because you need to reorder the attributes in `dimension.attributes`. Therefore it would be made useless if you had used an enum (you could directly iterate on the correct attribute order). Since you don't want to go that way, at least extract the reordering part from the name building logic to make this code simpler:
```rust
// you could write the "get attribute in an ordered fashion
// (or not)" in the CoverCrypt repository.
let attributes = dimension.get_attributes();
// The issue is that if the above function returns a `Vec`, you
// allocate twice and go through attributes twice. A solution
// could be to return an iterator on said attributes. This avoids
// both the double collecting and iteration issues. Maybe you
// could rename `get_attributes()` into `attributes()` to mimic the
// hash map's `.values()` method which also returns an iterator? 
let attributes = attributes.map(|(name, attr)| {
    // the rest goes unchanged
}).collect();
```

crate/cli/src/actions/cover_crypt/policy.rs
```diff
@@ -93,19 +91,39 @@ impl TryFrom<&Policy> for PolicySpecifications {
 
     fn try_from(policy: &Policy) -> Result<Self, Self::Error> {
         let mut result: HashMap<String, Vec<String>> = HashMap::new();
```
tbrezot, line 93: Allocate correct capacity.

crate/cli/src/actions/cover_crypt/policy.rs
```diff
@@ -93,19 +91,39 @@ impl TryFrom<&Policy> for PolicySpecifications {
 
     fn try_from(policy: &Policy) -> Result<Self, Self::Error> {
```
tbrezot, line 92: It seems you need to clone the `Policy` for this conversion (you are cloning many things out of it). If that is the case, the best would be to make it clear by taking ownership of it.

crate/kmip/src/kmip/kmip_operations.rs
```diff
@@ -1069,6 +1069,11 @@ pub struct ReKeyKeyPair {
     // Storage Mask selections that are permissible for the new Public Key object.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub public_protection_storage_masks: Option<ProtectionStorageMasks>,
+
+    // Specifies the action to perform during the rekeying
```
tbrezot, line 1073: Why not public documentation using `///`? It seems both the structure and the attribute are public.

crate/kmip/src/kmip/kmip_operations.rs
```diff
@@ -1069,6 +1069,11 @@ pub struct ReKeyKeyPair {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    // Defaults to RotateAttributes
```
tbrezot, line 1075: Defaults to nothing if `None`. The `Default` of `ReKeyKeyPairAction` is `RotateAttributes`, but this is a different matter. Why did you feel the need to specify it here?

crate/kmip/src/kmip/kmip_operations.rs
```diff
@@ -1080,6 +1085,19 @@ pub struct ReKeyKeyPairResponse {
     pub public_key_unique_identifier: UniqueIdentifier,
 }
 
+#[derive(Debug, Serialize, Deserialize, Default)]
+#[serde(rename_all = "PascalCase")]
+pub enum ReKeyKeyPairAction {
```
tbrezot, line 1090: Why not `RekeyKeyPairAction` (it seems to rekey [is a thing](https://wikiless.tiekoetter.com/wiki/Rekeying_(cryptography)?lang=en))?
```suggestion
pub enum RekeyKeyPairAction {
```

crate/kmip/src/kmip/kmip_operations.rs
```diff
@@ -1080,6 +1085,19 @@ pub struct ReKeyKeyPairResponse {
...
+    #[default]
+    RotateAttributes,
```
tbrezot, line 1092: Why is it a goof idea to make this action the default? Do you really need a default in the first place?

crate/pyo3/src/py_kms_client.rs
```diff
@@ -24,6 +27,47 @@ use rustls::Certificate;
...
+        $tags:expr,
+        $action:expr,
+        $py:ident
+    ) => {{
+        let id = match ($master_secret_key_identifier, $tags) {
+            (Some(key_id), _) => key_id,
+            (_, Some(tags)) => serde_json::to_string(&tags)
+                .map_err(|_e| PyException::new_err("invalid tag(s) specified"))?,
+            _ => return Err(PyException::new_err("please specify a key id or tags")),
+        };
```
tbrezot, line 46: Would it be a good idea to create a `Handle` type to avoid the `_` case?
```rust
pub enum KeyHandle {
    Id(String),
    Tags(Vec<Tag>),
}

\\ then...
let id = match handle {
    KeyHand::Id(id) => id,
    KeyHandle::Tags(tags) => serde_json::to_string(&tags).map_err(...))?,
};
```

You could even implement this as a method like `KeyHandle::try_to_string()`.

crate/pyo3/src/py_kms_client.rs
```diff
@@ -24,6 +27,47 @@ use rustls::Certificate;
...
+            .into_iter()
+            .map(Attribute::try_from)
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| PyTypeError::new_err(e.to_string()))?;
+
+        let request = build_rekey_keypair_request(&id, policy_attributes, $action)
+            .map_err(|e| PyException::new_err(e.to_string()))?;
+
+        let client = $self.0.clone();
+        pyo3_asyncio::tokio::future_into_py($py, async move {
```
tbrezot, line 58: You may want to wrap `self.0` into an `Arc` to clone it painlessly. Maybe it is not worth it though, depending on the cost of cloning `self.0` (difficult to judge from this piece of code).

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -138,6 +170,123 @@ fn rotate_policy(
...
+            .map_err(|e| {
+                KmsError::KmipError(
+                    ErrorReason::Unsupported_Cryptographic_Parameters,
+                    e.to_string(),
+                )
+            })?;
+    }
+    trace!("The new policy is : {policy:#?}");
+    Ok(policy)
+}
```
tbrezot, line 260: Here you need ownership on the input attributes so that you can add them to the policy.
```suggestion
fn add_policy_attribute(
    private_key: &Object,
    additions: Vec<(Attribute, EncryptionHint)>,
) -> KResult<Policy> {
    let private_key_attributes = private_key.attributes()?;
    let mut policy = policy_from_attributes(private_key_attributes)?;

    for (new_attribute, encryption_hint) in additions {
        policy
            .add_attribute(new_attribute, encryption_hint)
            .map_err(|e| {
                KmsError::KmipError(
                    ErrorReason::Unsupported_Cryptographic_Parameters,
                    e.to_string(),
                )
            })?;
    }
    trace!("The new policy is : {policy:#?}");
    Ok(policy)
}
```

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -138,6 +170,123 @@ fn rotate_policy(
     Ok(policy)
 }
 
+/// Removes old rotations from the policy of the given Master Private Key
+/// and return it
+fn clear_old_rotations_policy(
+    private_key: &Object,
+    cover_crypt_policy_attributes_to_update: &[Attribute],
+) -> KResult<Policy> {
```
tbrezot, line 178: After reading five times almost the same function, I wonder if it would not be better to take advantage of Rust's higher order functions:
```rust
fn mutate_policy_attributes(
    private_key: &Object,
    mutator: impl FnOnce(&mut Policy) -> KResult<()>,
) -> KResult<Policy> {
    let private_key_attributes = private_key.attributes()?;
    let mut policy = policy_from_attributes(private_key_attributes)?;
    mutator(&mut policy)?;
    Ok(policy)
}

fn add_policy_attribute(
    private_key: &Object,
    new_attributes: Vec<(Attribute, EncryptionHint)>,
) -> KResult<Policy> {
    mutate_policy_attributes(private_key, move |policy: &mut Policy| {
        for (attribute, encryption_hint) in new_attributes.into_iter() {
            policy.add_attribute(attribute, encryption_hint)?
        }
        Ok(())
    })
}

fn disable_policy_attributes(private_key: &Object, targets: &[Attribute]) -> KResult<Policy> {
    mutate_policy_attributes(private_key, move |policy: &mut Policy| {
        for attribute in targets {
            policy.disable_attribute(attribute.clone())?
        }
        Ok(())
    })
}
```

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -138,6 +170,123 @@ fn rotate_policy(
     Ok(policy)
 }
 
+/// Removes old rotations from the policy of the given Master Private Key
```
tbrezot, line 173: Same as a previous comment, I would rather like we make it clear that rotations associate values to attributes, thus the comment could be "removes the old attribute values from the policy".

crate/pyo3/src/py_kms_client.rs
ackRow: We tried to do it once but it implied too many changes in the rust KMS client.

crate/kmip/src/kmip/kmip_operations.rs
ackRow: in the previous versions, ReKeyKeyPair requests were only doing attribute rotations so this is the default behavior to keep backward compability.

crate/kmip/src/kmip/kmip_operations.rs
ackRow: I used the same naming convention as the other structs

crate/pyo3/src/py_kms_client.rs
```diff
@@ -24,6 +27,48 @@ use rustls::Certificate;
...
+        $tags:expr,
+        $policy_attributes:ident,
+        $action:expr,
+        $py:ident
+    ) => {{
+        let id = match ($master_secret_key_identifier, $tags) {
+            (Some(key_id), None) => key_id,
+            (None, Some(tags)) => serde_json::to_string(&tags)
+                .map_err(|_e| PyException::new_err("invalid tag(s) specified"))?,
+            _ => return Err(PyException::new_err("please specify a key id or tags")),
```
tbrezot, line 46: The error message does not match with the case `(Some(key_id), Some(tags))`

crate/pyo3/src/py_kms_client.rs
```diff
@@ -264,35 +309,168 @@ impl KmsClient {
...
+    ///
+    /// Returns:
+    ///     Future[Tuple[str, str]]: (Public key UID, Master secret key UID)
+    pub fn rename_cover_crypt_attribute<'p>(
+        &'p self,
+        attribute: &str,
+        new_name: &str,
+        master_secret_key_identifier: Option<String>,
+        tags: Option<Vec<&str>>,
+
```
tbrezot, line 457:
```suggestion
```

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -265,3 +220,70 @@ async fn refresh_all_user_decryption_keys(
...
+/// - `action`: An `EditPolicyAction` enum representing the action to perform on the policy.
+/// - `master_private_key`: A reference to the master private key object from which the policy
+///   attributes are retrieved.
+///
+/// # Returns
+///
+/// Returns a `Result` containing a tuple with the updated `Policy` and a vector of `Attribute`s
+/// that were modified during the action. If any errors occur during the policy updates or
+/// attribute retrieval, an `Err` variant is returned with a `KmsError`.
+fn dispatch_rekey_action(
```
tbrezot, line 242: The naming does not match the action imho
```suggestion
fn update_policy(
```

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -265,3 +220,70 @@ async fn refresh_all_user_decryption_keys(
 
     Ok(())
 }
+
+/// Update a Covercrypt policy based on the specified action and master private key.
```
tbrezot, line 224: It would be much clearer if you'd taken as argument a `Policy` instead of the master key. As of now, it is not clear form the documentation function whether or not the msk gets updated.

If you make this change, you may want to take ownership of the policy.

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -265,3 +220,70 @@ async fn refresh_all_user_decryption_keys(
...
+///
+/// - `action`: An `EditPolicyAction` enum representing the action to perform on the policy.
+/// - `master_private_key`: A reference to the master private key object from which the policy
+///   attributes are retrieved.
+///
+/// # Returns
+///
+/// Returns a `Result` containing a tuple with the updated `Policy` and a vector of `Attribute`s
+/// that were modified during the action. If any errors occur during the policy updates or
+/// attribute retrieval, an `Err` variant is returned with a `KmsError`.
```
tbrezot, line 241: Was this automatically generated? There is a lot of redundancy with function signature.
```suggestion
/// Returns the updated `Policy` along with the modified `Attribute`s.
```

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -265,3 +220,70 @@ async fn refresh_all_user_decryption_keys(
...
+/// This function takes an `EditPolicyAction` and a master private key as input and performs the
+/// corresponding policy updates. It retrieves the current policy from the master private key's
+/// attributes and applies the requested action, resulting in a new policy and a list of attributes
+/// that were updated or modified.
+///
+/// # Parameters
+///
+/// - `action`: An `EditPolicyAction` enum representing the action to perform on the policy.
+/// - `master_private_key`: A reference to the master private key object from which the policy
+///   attributes are retrieved.
```
tbrezot, line 235: Avoid redundancy with function signature
```suggestion
/// - `action`: action to perform on the policy.
/// - `master_private_key`: master private key for which the policy is updated.
```

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -265,3 +220,70 @@ async fn refresh_all_user_decryption_keys(
...
+            ErrorReason::Unsupported_Cryptographic_Parameters,
+            e.to_string(),
+        )
+    })?;
+
+    let attributes_to_update = match action {
+        EditPolicyAction::RotateAttributes(attrs) => attrs,
+        EditPolicyAction::ClearOldRotations(attrs) => attrs,
+        _ => vec![], // no need to update existing usk
+    };
```
tbrezot, line 285: Several points:
1. the naming makes it unclear whether some more action needs to be performed on these attributes;
2. the naming is not coherent with the function doc which clearly states that an action has been performed on these attributes;
3. you may not need this second match (which would avoid a clone in the `EditPolicyAction::AddAttribute` case.

crate/server/src/core/cover_crypt/update_policy.rs
```diff
@@ -28,31 +28,12 @@ pub async fn rekey_keypair_cover_crypt(
     kmip_server: &KMS,
     cover_crypt: Covercrypt,
```
tbrezot, line 29: Should you really take ownership of this?

crate/utils/src/crypto/cover_crypt/attributes.rs
```diff
@@ -239,3 +241,51 @@ pub fn header_uid_from_attributes(attributes: &Attributes) -> Result<&[u8], Kmip
...
+                format!("failed serializing the CoverCrypt action: {e}"),
+            )
+        })?,
+    })
+}
+
+/// Extract an edit `CoverCrypt` policy action from attributes
+pub fn edit_policy_action_from_attributes(
+    attributes: &Attributes,
+) -> Result<EditPolicyAction, KmipError> {
```
tbrezot, line 274: Why didn't you take a `VendorAttribute` as argument? This would make it symmetrical with the reverse operation. Plus it seems you are ignoring the error raised here in the caller (`rekey_keypair:85`)... which means that an error to parse the operation (let's say a user misspell) will result in a rotation... 

If you want to keep it here, maybe can you at least differentiate internal errors and default to a rotation only when no vendor attribute is passed?

tbrezot (commented):
Almost good to go :)

crate/server/src/core/cover_crypt/update_policy.rs
ackRow: Changed but it seems that the KMS is designed to create a new Covercrypt object at each call:
```rust
rekey_keypair_cover_crypt(kms, Covercrypt::default(), &owm.id, user, action, params).await
```

tbrezot: Yes, you can raise an issue. It may be better to change this.

PR #70, opened 2023-10-19
Bulk operations for Symmetric keys and Elliptic Curves keys

### LINKED ISSUE #82: Generalize bulk operation using KMIP Messages struct

ThibsG:
fixes #82 

#37  introduced the bulk encryption/decryption for Covercrypt, by defining a new cryptographic algorithm enum variant and serializing multiple data into only one vector of bytes. The goal was to be able to perform multiple operations with only one request to the server and grab some bytes for optimization.

We could implement the same scheme for Symmetric/EC keys, but it would need to duplicate each cryptographic algorithm enum variant for each algorithm we want to support.

KMIP introduced a way to ask for multiple operations in only one request to the server: KMIP Messages.

One KMIP message could include several 'batch item' to be processed by the server.
Each item is completely independent from the others.

The adding of KMIP Messages involves the following:
- [X] create structs and enums of Message and sub items (and their (de)serialization)
- [X] unroll batch items and start their processing
- [X] make the whole thing accessible through REST API
- ~~[X] make the whole thing accessible through CLI?~~ 

see "8. Messages": https://docs.oasis-open.org/kmip/kmip-spec/v2.1/os/kmip-spec-v2.1-os.html


crate/server/src/tests/kmip_messages.rs
```diff
@@ -0,0 +1,136 @@
...
+    // prepare and send the single message
+    let items = vec![
+        MessageBatchItem {
+            operation: OperationEnumeration::CreateKeyPair,
+            ephemeral: None,
+            unique_batch_item_id: None,
+            request_payload: Operation::CreateKeyPair(ec_create_request),
+            message_extension: None,
+        },
+        MessageBatchItem {
```
Manuthor, line 38: implement `Default`?

crate/server/src/tests/cover_crypt_tests/integration_tests_bulk.rs
```diff
@@ -0,0 +1,110 @@
...
+                    &policy, EMPTY_TAGS,
+                )?),
+                message_extension: None,
+            },
+        ],
+    };
+
+    let response: MessageResponse = test_utils::post(&app, &request_message).await?;
+
+    // tracing::trace!("{response:#?}");
```
Manuthor, line 80: leftover?

Manuthor (commented):
great job thank you!

I wonder if you should keep the `CryptographicAlgorithm::CoverCryptBulk`.
Is there any impact on the doc?

Manuthor (approved):

PR #89, opened 2023-11-13
test: check with cloudproof_kms_js

Manuthor:


crate/server/src/core/operations/import.rs
```diff
@@ -241,31 +239,32 @@ async fn create_certificate_link(
...
-    pem_value: &[u8],
+    der_value: &[u8],
     kms: &KMS,
     owner: &str,
     params: Option<&ExtraDatabaseParams>,
 ) -> KResult<Object> {
-    let (_, pem) = parse_x509_pem(pem_value)?;
+    let result = parse_x509_certificate(der_value);
+
+    let object = if result.is_ok() {
```
ThibsG, line 252: Just a suggestion, maybe something like this is more explicit:
```suggestion
    let der_is_a_cert = parse_x509_certificate(der_value).is_ok();

    let object = if der_is_a_cert {
```

if you don't like `der_is_a_cert`, maybe `cert_parsing_is_ok`?

ThibsG (approved):

PR #99, opened 2023-11-24
docs: fix invalid links

Manuthor:


Dockerfile
```diff
@@ -41,6 +46,13 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 COPY --from=builder /root/kms/target/release/cosmian_kms_server /usr/bin/cosmian_kms_server
+COPY --from=builder /root/kms/target/release/ckms               /usr/bin/ckms
+
+#
+# Create working directory
+#
+RUN mkdir /data
+WORKDIR /data
```
ThibsG, line 55: `WORKDIR` cmd already creates the folder if it doesn't exist
```suggestion
WORKDIR /data
```

delivery/docker-compose.yml
```diff
@@ -1,23 +0,0 @@
...
-  kms:
-    container_name: kms
-    image: kms
-    environment:
-      - KMS_POSTGRES_URL=postgres://kms:kms@db/kms
-      - KMS_HOSTNAME=0.0.0.0
-    ports:
-      - 9998:9998
-    depends_on:
-      - db
```
ThibsG, line 23: maybe we should move this part into the `docker-compose.yml` file at root of the project, what do you think?

ThibsG (approved):
nice cleanup,, thanks!

in root `README.md`, there is a reference to the folder `delivery` you might want to remove also

Dockerfile
Manuthor: right

delivery/docker-compose.yml
Manuthor: well if you don't mind I would rather keep the KMS service outside (the docker-compose has only the KMS external services)

PR #127, opened 2023-12-21
Config file

heavenboy8:


README.md
```diff
@@ -70,3 +70,11 @@ supervisorctl reload
...
 supervisorctl status cosmian_kms
 ```
+
+## Server parameters
+
+Order of variables for server configuration:
+- args in command line
+- env var
+- conf file
+- default (set on struct)
```
ThibsG, line 80: not sure this is true with the latest impl, what do you think?

ThibsG (approved):

README.md
ThibsG: see https://github.com/Cosmian/kms/pull/127/commits/fa4f9a4b2fabf0a28a55ce87dffe7725c9f3f5ec

PR #128, opened 2023-12-21
feat: dynamic salt for password derivation, resolving issue #124

### LINKED ISSUE #124: Critical: better salt for password derivation
Passwords can be derived into secure keys via the `derive_key_from_password()` method in `crate/utils`.
There is an issue where one would not find back the same key from a password twice and because the salt is not saved.
This could be fixed by adding an `Option(&[u8])` parameter to `derive_key_from_password()` to both FIPS (PBKDF2 with HMAC256) and non-FIPS (Argon2).

Adding support for non-deterministic salt to Argon2 is primordial since for the moment the salt is a constant string, which means that two identical password derive to the same key and this need to be avoided.

Don't forget to ad tests to `derive_key_from_password()` with optional salt.

JosePisco:
Regarding #124 , I fixed the issue of dynamic salting for password derivation.

I wrote some remarks, I advise you read, regarding the strategy we should use for this (user or server salt handling).

Closes #124


crate/cli/Cargo.toml
```diff
@@ -24,6 +24,7 @@ cosmian_kms_client = { path = "../client" }
 cosmian_kms_utils = { path = "../utils" }
 cosmian_logger = { path = "../logger" }
 der = { workspace = true }
+hex = "0.4"
```
Manuthor, line 27: is hex encoding is really required? Can input salt be encoded in base64 instead?

crate/cli/src/actions/shared/unwrap_key.rs
```diff
@@ -49,6 +49,16 @@ pub struct UnwrapKeyAction {
     )]
     unwrap_password: Option<String>,
 
+    /// An optional hexadecimal salt to derivate the password into a key.
+    #[clap(
+        long = "passwd-salt",
```
Manuthor, line 54:
```suggestion
        long = "password-salt",
```

crate/server/src/config/command_line/db.rs
```diff
@@ -177,8 +177,9 @@ impl DBConfig {
                         "KMS_REDIS_MASTER_PASSWORD",
                     )?;
                     // Generate the symmetric key from the master password
-                    let master_key =
-                        RedisWithFindex::master_key_from_password(&redis_master_password)?;
+                    // TODO - store salt somewhere ?
```
Manuthor, line 180: IMHO, KMS server should not handle salts for customers unless this feature is specifically required. Customers should be informed as soon as possible (at the top level of salt use, e.g in the CLI) that provided salts are (or are not) FIPS compliant

crate/utils/src/crypto/cover_crypt/kmip_requests.rs
```diff
@@ -95,8 +95,9 @@ pub fn build_import_decryption_private_key_request<T: IntoIterator<Item = impl A
     //  - to wrapped (wrapping_password is some)
     //  - or not wrapped (otherwise)
     let is_wrapped = is_wrapped || wrapping_password.is_some();
+    // TODO - is salt necessary here ?
```
Manuthor, line 98: since wrapping_password could be weak, salt should be necessary here. Probably create an issue with low priority.

Manuthor (commented):
LGTM

crate/cli/Cargo.toml
JosePisco: Yes sure, commit f152a523ae5f3070783edd03674145a6b82b9f44 should fix this.

crate/cli/src/actions/shared/unwrap_key.rs
JosePisco: commit 8c674423862328a80812281746a4ff4a9cb50926 should fix this.

crate/utils/src/crypto/cover_crypt/kmip_requests.rs
JosePisco: If I understand correctly this part of the code, the database master key is created upon database initialization and there should be no use of needing to compute it again ? If so, then there is no point in storing the salt. What do you think ?

crate/cli/src/actions/shared/wrap_key.rs
```diff
@@ -75,12 +75,19 @@ impl WrapKeyAction {
...
             let key_bytes =
-                derive_key_from_password::<SYMMETRIC_WRAPPING_KEY_SIZE>(password.as_bytes())?
-                    .to_vec();
-            create_symmetric_key(&key_bytes, CryptographicAlgorithm::AES)
+                derive_key_from_password::<SYMMETRIC_WRAPPING_KEY_SIZE>(password.as_bytes())?;
+
+            let symmetric_key_object =
+                create_symmetric_key(&key_bytes, CryptographicAlgorithm::AES)?;
+
+            println!(
```
JosePisco, line 86: is this println! still needed?

crate/cli/src/actions/shared/unwrap_key.rs
```diff
@@ -91,12 +78,7 @@ impl UnwrapKeyAction {
             let key_bytes = general_purpose::STANDARD
                 .decode(b64)
                 .with_context(|| "failed decoding the unwrap key")?;
-            create_symmetric_key(&key_bytes, CryptographicAlgorithm::AES)
```
JosePisco, line 94: I'd drop a § of comments to explain why unwrap is not symmetrical to wrap and why we removed the password option. Or where it fits best in code.

crate/cli/src/actions/shared/wrap_key.rs
Manuthor: this one is required (since this is the only time that the key will be printed)

Manuthor: adding a comment

crate/cli/src/actions/shared/unwrap_key.rs
Manuthor: right

PR #130, opened 2023-12-22
112 make rsa oaep aes a generalized encryption system for use in all kms and not only for key wrapping

JosePisco:


crate/utils/src/crypto/wrap/encrypt_decrypt.rs
```diff
@@ -176,47 +167,37 @@ pub fn decrypt_bytes(
...
 
-#[cfg(feature = "fips")]
 fn decrypt_with_private_key(
     p_key: PKey<Private>,
     ciphertext: &[u8],
 ) -> Result<Vec<u8>, KmipUtilsError> {
-    // Unwrap symmetric key using RSA.
-    if p_key.rsa().is_err() {
-        kmip_utils_bail!("Decryption Error: Only RSA KEM supported in FIPS mode.")
+    if p_key.rsa().is_ok() {
```
Manuthor, line 174: looks like we are still using `p_key` even if `is_ok()` is false. Since this code is not FIPS related anymore, can it be merge?

crate/utils/src/crypto/hybrid_encryption/rsa_oaep_aes_gcm.rs
```diff
@@ -0,0 +1,174 @@
...
+            FIPS_MIN_RSA_MODULUS_LENGTH,
+            pubkey.bits()
+        )
+    }
+
+    let encapsulation_bytes_len = rsa_pubkey.size() as usize;
+
+    // Generate temporary AES key.
+    let mut kek = Zeroizing::from(vec![0u8; AES_256_GCM_KEY_LENGTH]);
+    rand_bytes(&mut kek)?;
```
Manuthor, line 51: Just a general comment: at one point, we should not depend on crypto_core anymore (except for Findex): since Openssl do not provide zeroize implementation, IMHO we should use directly zeroize.

Manuthor (commented):
LGTM

crate/utils/src/crypto/hybrid_encryption/rsa_oaep_aes_gcm.rs
```diff
@@ -0,0 +1,174 @@
...
+};
+
+use crate::{
+    crypto::symmetric::{AES_256_GCM_IV_LENGTH, AES_256_GCM_KEY_LENGTH, AES_256_GCM_MAC_LENGTH},
+    error::KmipUtilsError,
+    kmip_utils_bail,
+};
+
+#[cfg(feature = "fips")]
+const FIPS_MIN_RSA_MODULUS_LENGTH: u32 = 256;
```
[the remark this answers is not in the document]
JosePisco: rust-openssl does not expose a lot of constants like key size, privkey size, etc. unfortunately.

crate/utils/src/crypto/wrap/encrypt_decrypt.rs
JosePisco: Indeed, commit 7e5e9403330e8683c2bbddc9f5f4864a4bfac9ed should fix this.

crate/utils/src/crypto/hybrid_encryption/rsa_oaep_aes_gcm.rs
JosePisco: So instead of importing `cryptocore::reexport::zeroize` we better import `zeroize` directly ?
Maybe consider doing this in another issue ?

crate/utils/src/crypto/hybrid_encryption/encryption.rs
```diff
@@ -69,21 +68,21 @@ impl HybridEncryptionSystem {
...
-            )
-        }
+        let mut encrypt_response = EncryptResponse {
+            unique_identifier: UniqueIdentifier::TextString(self.public_key_uid.clone()),
+            data: None,
+            iv_counter_nonce: None,
+            correlation_value: None,
+            authenticated_encryption_tag: None,
+        };
+
```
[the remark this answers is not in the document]
JosePisco: commit https://github.com/Cosmian/kms/commit/7e5e9403330e8683c2bbddc9f5f4864a4bfac9ed should fix this.

PR #133, opened 2024-01-04
Fix AES decryption

### LINKED ISSUE #131: Fix Panic if AuthenticatedEncryptionTag is missing

ThibsG:
#131 was due to a bad check of input data [here](https://github.com/Cosmian/kms/blob/b3c66863feb923be91d1b33f0d7570b04ce2f2bd/crate/utils/src/crypto/symmetric/aes_256_gcm.rs#L169) when [creating tag](https://github.com/Cosmian/kms/blob/b3c66863feb923be91d1b33f0d7570b04ce2f2bd/crate/utils/src/crypto/symmetric/aes_256_gcm.rs#L150) (bad default). The tag was then used in [`GenericArray::from`](https://github.com/Cosmian/crypto_core/blob/4b337535e683ccf475bcf6b79c58c421cf272e3d/src/symmetric_crypto/dem.rs#L136) and as `tag` could be empty, it was panicking because of invalid size.

This encrypt/decrypt part has been replaced since the issue by the use of openssl, but the `tag` can still fail at conversion, due to empty vector not fitting 16 bytes required size.

the real fix in this PR:
```diff
        // Recover tag. Ensure it is of correct size.
        let tag: [u8; AES_256_GCM_MAC_LENGTH] = request
            .authenticated_encryption_tag
            .clone()
-            .unwrap_or_default()
+            .unwrap_or(vec![0_u8; AES_256_GCM_MAC_LENGTH])
            .try_into()?;
```

fixes #131

### DIFF
--- a/crate/utils/src/crypto/symmetric/aes_256_gcm.rs
+++ b/crate/utils/src/crypto/symmetric/aes_256_gcm.rs
@@ -26,7 +26,7 @@ pub struct AesGcmSystem {
 impl AesGcmSystem {
     pub fn instantiate(uid: &str, symmetric_key: &Object) -> Result<Self, KmipUtilsError> {
         let key_block = match symmetric_key {
-            Object::SymmetricKey { key_block } => key_block.clone(),
+            Object::SymmetricKey { key_block } => key_block,
             _ => {
                 return Err(KmipUtilsError::NotSupported(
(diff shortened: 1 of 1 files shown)


JosePisco (approved):
Looks good to me.

PR #138, opened 2024-01-08
Set commands in a separate public module for kms_gui repository

ThibsG:
related to the work of: https://github.com/Cosmian/kms_gui


ThibsG:
Finally, `kms_gui` is using a different approach to use the KMS CLI, so no need to make these items `pub`

PR #160, opened 2024-01-29
Dynamic tmp directory for tests

JosePisco:
Tests used to write files in /tmp with the same filename which caused multi-threaded testing to induce race conditions sometimes.

These changes aim to create dedicated tmp directories for each test. This cancels concurrent read/write operations from two different tests and should fix the random fails in the testsuite.


ThibsG (approved):
Thanks! 🚀

Issue #159, opened 2024-01-26
Zeroize BigUInt keys


tbrezot:
If I am right, you are using BigUint for storage only. Therefore it should be possible to implement the correct operation on dropping structure that own them (big integers are essentially pointers so they do not leak there content in the memory when they are passed around).

_Originally posted by @tbrezot in https://github.com/Cosmian/kms/pull/149#discussion_r1459156583_

JosePisco:
`DecryptionResponse` struct contains decrypted plaintext which also needs to be zeroized in file `crate/kmip/src/kmip/kmip_operations.rs`.

Issue #136, opened 2024-01-04
Zeroize crypto


JosePisco:
At this point, many processed data in cryptographic primitives are not properly zeroized.
I propose to use the `zeroize` crate to create `Zeroizing` objects from keys and plaintexts.

Make sure to use the `zeroize` crate directly if it is imported through cloudproof.

PR #164, opened 2024-01-31
Better zeroization for fixed size keys and removing non-fips rng from server

### LINKED ISSUE #106: FIPS version of the Server
Create a special `FIPS` version of the server that will only enable FIPS-140 certified crypto modules.

The choice is to use OpenSSL (only) server side in one of its FIPS-140 approved modes.

### LINKED ISSUE #153: Change Key types from Vec<u8> to some Secret custom type
crypto_core [proposes a new `Secret` type](https://github.com/Cosmian/crypto_core/pull/74) that allocates arrays of bytes on the heap zeroizes on drop.

It would be clearer both for the architecture to force Symmetric keys back to fixed length array. But because arrays are stack-allocated, they are a pain to zeroize properly. This feature could be a solution to our problems.

JosePisco:
This pull requests enhances zeroization onSecret keys of fixed size by wrapping them in a structure called `Secret` which is a Pin Box on the heap that is easier to track and zeroize. This structure was added in `crypto_core` by @tbrezot [here](https://github.com/Cosmian/crypto_core/pull/74).

This PR also reviewed the kms server and changed any non-zeroized symmetric keys into `Secret` types. It also gets rid of any non-fips rng sources **which means that the rng field in the KMS structure has been removed**.

It is important to note that `findex_with_redis` remains with old non-zeroized symmetric keys.

This should solve #153 and #106. Remains #159 to fully solve #136.

Closes #153 
Closes #106


Manuthor (approved):
LGTM, thanks

crate/utils/src/access.rs
```diff
@@ -91,16 +91,15 @@ impl<'de> Deserialize<'de> for ExtraDatabaseParams {
...
         let group_id_bytes: [u8; 16] = bytes[0..16]
             .try_into()
             .map_err(|_| serde::de::Error::custom("Could not deserialize ExtraDatabaseParams"))?;
         let group_id = u128::from_be_bytes(group_id_bytes);
         let key_bytes: [u8; 32] = bytes[16..48]
             .try_into()
             .map_err(|_| serde::de::Error::custom("Could not deserialize ExtraDatabaseParams"))?;
-        let key = SymmetricKey::try_from_bytes(key_bytes)
-            .map_err(|_| serde::de::Error::custom("Could not deserialize ExtraDatabaseParams"))?;
+        let key = Secret::from_protected_bytes(&key_bytes);
```
tbrezot, line 102: Why not calling `from_unprotected_bytes` directly on the slice `&bytes[16..48]` ? You would avoid a copy.

crate/utils/src/crypto/password_derivation.rs
```diff
@@ -39,7 +39,8 @@ pub fn derive_key_from_password<const LENGTH: usize>(
         )
     }
 
-    let mut output_key_material = Zeroizing::from(vec![0u8; LENGTH]);
+    //let mut output_key_material = Zeroizing::from(vec![0u8; LENGTH]);
```
tbrezot: dead code
```suggestion
```

JosePisco: this is fixed in last commit

crate/utils/src/access.rs
JosePisco: Because bytes are already zeroized.

crate/cli/src/actions/shared/wrap_key.rs
```diff
@@ -87,7 +87,7 @@ impl WrapKeyAction {
             // Print the wrapping key for user. This is the only time that this wrapping key will be printed
             println!(
                 "Wrapping key: {}",
-                general_purpose::STANDARD.encode(key_bytes)
+                general_purpose::STANDARD.encode(key_bytes.deref())
```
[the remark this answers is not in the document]
JosePisco: Your suggestion does not work as the Trait AsRef<[u8]> is not implemented for Secret.
I suggest to leave it like this instead of writing more code, especially since the end goal is the same ?

crate/server/src/core/implementation.rs
```diff
@@ -87,9 +79,9 @@ impl KMS {
                     // So we are going to create a "zeroizable" copy which will be passed to Redis with Findex
                     // and zerorize the one in the shared config
                     let new_master_key =
-                        SymmetricKey::<REDIS_WITH_FINDEX_MASTER_KEY_LENGTH>::try_from_bytes(
-                            master_key.to_bytes(),
-                        )?;
+                        Secret::<REDIS_WITH_FINDEX_MASTER_KEY_LENGTH>::from_unprotected_bytes(
+                            &mut master_key.to_bytes(),
+                        );
                     master_key.zeroize();
```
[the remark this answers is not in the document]
JosePisco: I agree an unnecessary copy is done but I encountered problems making `Secret` Understand `SymmetricKey` is just an array of bytes. Your suggestion does not work tho, it recognizes only a `&mut` object

PR #165, opened 2024-01-31
Refactor utils

ThibsG:
- move `utils/crypto` into `kmip/crypto`
- dispatch other stuff in server, kmip and client
- remove `utils` subcrate


ThibsG:
superseeded by #178

PR #166, opened 2024-02-01
feat: save KMIP attributes next to KMIP-object

### LINKED ISSUE #88: Improve KMIP `Attributes` support
Kmip managed objects (`Objects`) are associated with attributes ( `Attributes`).
These attributes contain meta information on the object, in particular links to other objects.

To store the attributes, the current implementation relies on a nested `Attributes` property inside the [`KeyValue`](https://docs.oasis-open.org/kmip/kmip-spec/v2.1/os/kmip-spec-v2.1-os.html#_Toc57115548), located inside a `KeyBlock` which is itself inside the `Object`.

This implementation is problematic for a few reasons:

- Some `Object`s such as `Certificate` do not have a nested `Attributes` property where to store them, and attributes are lost on Import or not generated on Create.  The current implementation partially compensates with the use of tags, but this is limited and non-standard. 
- The current implementation prevents the use of the `Locate` operation to find `Object` that is not holding attributes.
- With regards to the `Attributes` nested inside the `KeyValue`, KMIP specifies they should merely be a copy of the "main" attributes

> This attribute information differs from the attributes associated with Managed Objects, and is obtained via the Get Attributes operation, only by the fact that it is encapsulated with (and possibly wrapped with) the key material itself.

### Proposed change

This ticket proposes to change the storage layer so that it stores the tuple `(Object, Attributes)` rather than just the `Object`. This would affect the `Database` trait in `crate/server/src/database/database_trait.rs` and its specific DB implementations.

The `Locate` operation should be updated to search these top-level Attributes.
The `Import` operation should be updated accordingly.

Manuthor:
Fixes #88 by creating the new JSON-column `Attributes` in the `objects` table:

- [x] Save KMIP Attributes in a proper column of `Objects` table:
  - [x] Remove all custom tags `_cert_spki`, `_cert_cn`, `_cert_issuer` and `_cert_sk`
  - [x] Pgsql
  - [x] Mysql
  - [x] Sql cipher
  - [x] Redis with findex: but no KMIP-Attributes saved (done on purpose)


crate/cli/src/actions/certificates/certify.rs
```diff
@@ -65,7 +65,7 @@ pub struct CertifyAction {
     #[clap(long = "days", short = 'd', default_value = "365")]
     number_of_days: usize,
 
-    /// The path to a X509 extension's file, containing a `v3_ca` parag
+    /// The path to a X509 extension's file, containing a `v3_ca` param
```
ThibsG, line 68: I really meant `parag`, a short for paragraph

crate/server/src/core/operations/certify.rs
```diff
@@ -174,13 +175,15 @@ pub async fn certify(
...
+                    attributes.clone(),
                     Some(tags),
                     StateEnumeration::Active,
                 )),
                 // update the public key
                 AtomicOperation::UpdateObject((
                     public_key_owm.id.clone(),
-                    public_key_owm.object,
+                    public_key_owm.object.clone(),
+                    public_key_owm.object.attributes()?.clone(),
```
ThibsG, line 186: avoid to clone the whole object by cloning the attributes first:
```rust
        let pk_own_obj_attributes = public_key_owm.object.attributes()?.clone();
        // return
        (
            issued_certificate_id.clone(),
            vec![
                // upsert the certificate
                AtomicOperation::Upsert((
                    issued_certificate_id,
                    issued_certificate,
                    attributes.clone(),
                    Some(tags),
                    StateEnumeration::Active,
                )),
                // update the public key
                AtomicOperation::UpdateObject((
                    public_key_owm.id.clone(),
                    public_key_owm.object,
                    pk_own_obj_attributes,
```

crate/server/src/core/operations/create_key_pair.rs
```diff
@@ -28,24 +28,22 @@ pub async fn create_key_pair(
...
+            key_pair.private_key().clone(),
+            key_pair.private_key().attributes()?.clone(),
+            sk_tags,
+        )),
+        AtomicOperation::Create((
+            pk_uid.clone(),
+            key_pair.public_key().clone(),
+            key_pair.public_key().attributes()?.clone(),
+            pk_tags,
+        )),
```
ThibsG, line 44: same here

crate/server/src/core/operations/export_utils.rs
```diff
@@ -632,7 +632,7 @@ async fn process_certificate(
     params: Option<&ExtraDatabaseParams>,
 ) -> KResult<Attributes> {
     //TODO: create attributes from tags until https://github.com/Cosmian/kms/issues/88 is fixed
-    let mut attributes = Attributes::default();
+    let mut attributes = owm.attributes.clone();
```
ThibsG, line 635: so... is the comment still accurate after this PR get merged?

crate/server/src/core/operations/import.rs
```diff
@@ -222,7 +241,13 @@ async fn process_public_key(
     let replace_existing = request.replace_existing.unwrap_or(false);
     Ok((
         uid.clone(),
-        vec![single_operation(tags, replace_existing, object, uid)],
+        vec![single_operation(
+            tags,
+            replace_existing,
+            object.clone(),
+            object.attributes()?.clone(),
```
ThibsG, line 248: same here

crate/server/src/core/operations/import.rs
```diff
@@ -281,7 +306,13 @@ async fn process_private_key(
 
         return Ok((
             uid.clone(),
-            vec![single_operation(tags, replace_existing, object, uid)],
+            vec![single_operation(
+                tags,
+                replace_existing,
+                object.clone(),
+                object.attributes()?.clone(),
```
ThibsG, line 313: same here

crate/server/src/core/operations/import.rs
```diff
@@ -305,12 +336,18 @@ async fn process_private_key(
...
         request.unique_identifier.as_str().unwrap_or_default(),
     )?;
     Ok((
         sk_uid.clone(),
-        vec![single_operation(sk_tags, replace_existing, sk, sk_uid)],
+        vec![single_operation(
+            sk_tags,
+            replace_existing,
+            sk.clone(),
+            sk.attributes()?.clone(),
```
ThibsG, line 348: same here

crate/server/src/core/operations/import.rs
```diff
@@ -451,22 +500,23 @@ async fn process_pkcs12(
...
+            // for PKCS12CertificateLink
+            LinkType::PKCS12CertificateLink,
+            LinkedObjectIdentifier::TextString(leaf_certificate_uid.clone()),
+        );
     operations.push(single_operation(
         private_key_tags,
         replace_existing,
-        private_key,
+        private_key.clone(),
+        private_key.attributes()?.clone(),
```
ThibsG, line 519: same here

documentation/docs/cli/main_commands.md
```diff
@@ -582,7 +592,7 @@ Possible values:  `"pem", "der"` [default: `"pem"`]
 
 `--days [-d] <NUMBER_OF_DAYS>` The requested number of validity days The server may grant a different value
 
-`--certificate-extensions [-e] <CERTIFICATE_EXTENSIONS>` The path to a X509 extension's file, containing a `v3_ca` parag
+`--certificate-extensions [-e] <CERTIFICATE_EXTENSIONS>` The path to a X509 extension's file, containing a `v3_ca` param
```
ThibsG, line 595: need to regenerate it

crate/cli/src/actions/certificates/certify.rs
Manuthor: eventually, could we change it to `v3_ca section`?

crate/server/src/core/operations/certify.rs
Manuthor: thanks!

crate/cli/src/actions/certificates/certify.rs
Manuthor: reverted

crate/cli/src/actions/shared/locate.rs
```diff
@@ -121,22 +111,7 @@ impl LocateObjectsAction {
...
-            tags.push(format!(
-                "_cert_spki={}",
-                certificate_spki.replace(':', "").to_lowercase()
-            ));
-        }
-
-        if let Some(tags) = opt_tags {
+        if let Some(tags) = self.tags.clone() {
             set_tags(&mut attributes, tags)?;
         }
```
ThibsG, line 116:
```suggestion
        if let Some(tags) = &self.tags {
            set_tags(&mut attributes, tags.clone())?;
        }
```

crate/server/src/core/operations/get_attributes.rs
```diff
@@ -71,11 +70,8 @@ pub async fn get_attributes(
 
     let attributes = match &owm.object {
         Object::Certificate { .. } => {
-            let mut attributes = Attributes::default();
-            add_certificate_tags_to_attributes(&mut attributes, &owm.id, kms, params).await?;
-            attributes.key_format_type = Some(KeyFormatType::X509);
-            attributes.object_type = Some(ObjectType::Certificate);
-            attributes
+            // KMIP Attributes retrieved from dedicated column `Attributes``
```
ThibsG, line 73:
```suggestion
            // KMIP Attributes retrieved from dedicated column `Attributes`
```

ThibsG (approved):

PR #173, opened 2024-02-05
CI improvements

ThibsG:
- remove large files (> 40 MB) before saving cache
- stop compiling for running tests on binaries (0 tests to pass, but still 4 mins to compile)
- unify binary compilations in a single line by using `--bins`

> #### CI total time
> before: 41 mins
> after: 31 mins

- stop doing release jobs at every push. Now a nightly pipeline will do the release jobs for develop (aka `latest_build` release packages). Tags still trigger a release build
> #### CI total time
> before: 31 mins
> after: 10 mins


ThibsG:
> Great job! It might come in very handy!
> 
> Also, if I understand the binary build, we don't strip the binary. We can try to set the `strip` flag to `true`, it lightens the produced binary by removing all unnecessary symbols and debug informations.
> 
> ```
> [profile.release]
> strip = "debuginfo"
> ```
> 
> Furthermore, I think we can activate LTOs (Link Time Optimizations) in rust to lighten the binary even more.
> 
> Quickly, LTOs allow the compiler to keep functions information aside and the linker can optimize even more thanks to those informations becaue he's got all final compiled files to deal with. In the cases that some short functions are in libraries (which might often be the case in all the externals crates we use), the compiler will never inline them because they need to be used by the binary, but the linker can see it when it's got all compiled files, and it can inline it in the end. It's useful to get rid of some more symbols but we need to know that it actually slows down the compile process by some important factor.

Hey thanks, you are totally right 😉 The binaries are already stripped and LTO is already activated, you can see this in the [workspace's `Cargo.toml` file](https://github.com/Cosmian/kms/blob/develop/Cargo.toml#L17):
```toml
[profile.release]
lto = true
strip = true

[profile.dev]
strip = "debuginfo"
incremental = false
```

PR #184, opened 2024-02-15
Fix interpolation in error macros

ThibsG:
This PR fixes string interpolation.

Before, a code like:
```rust
let var = 42;
let err = kms_error!("interpolate {var}");
```
was rendering
```
"interpolate {var}"
```
with this PR it is rendering

```
"interpolate 42"
```

Also, this PR [fixes](https://github.com/Cosmian/kms/pull/184/commits/5b24cf2d0d5a77abf17b8ea308fe6af42f823abb#diff-460ab4609eec2aef5e10adf4fe17629e6cdf5a3cfd952de5a7256022006d4da5R61) the `Secret` instantiation by turning data init from stack to heap

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -17,7 +17,7 @@ jobs:
       archive-name: centos7_tests
       commands: |
         cargo build --bins
-        cargo test -- --nocapture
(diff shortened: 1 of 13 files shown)


Manuthor (approved):
well done!

PR #185, opened 2024-02-16
Make KMIP subcrate compile without any activated feature

ThibsG:
```bash
$ cargo build -p cosmian_kmip
```
🎉

### DIFF
--- a/crate/kmip/Cargo.toml
+++ b/crate/kmip/Cargo.toml
@@ -10,7 +10,7 @@ license-file = "../../LICENSE.md"
 doctest = false
 
 [features]
-default = []
+default = ["dep:sha3"]
 openssl = ["dep:openssl"]
 pyo3 = ["dep:pyo3"]
 # Enable FIPS module feature build. KMS builds in FIPS mode when this is enabled.
--- a/crate/kmip/src/crypto/mod.rs
+++ b/crate/kmip/src/crypto/mod.rs
@@ -2,10 +2,13 @@ pub mod cover_crypt;
 pub mod dh_shared_keys;
 pub mod elliptic_curves;
 pub mod generic;
+#[cfg(any(feature = "openssl", feature = "fips"))]
 pub mod password_derivation;
 pub mod rsa;
 pub mod secret;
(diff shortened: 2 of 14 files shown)


crate/kmip/src/crypto/mod.rs
```diff
@@ -2,10 +2,13 @@ pub mod cover_crypt;
 pub mod dh_shared_keys;
 pub mod elliptic_curves;
 pub mod generic;
+#[cfg(any(feature = "openssl", feature = "fips"))]
```
JosePisco, line 5: What if not openssl and not fips ?
There should be a way to derive with Argon2 that does not need fips nor openssl.

crate/kmip/Cargo.toml
```diff
@@ -10,7 +10,7 @@ license-file = "../../LICENSE.md"
 doctest = false
 
 [features]
-default = []
+default = ["dep:sha3"]
 openssl = ["dep:openssl"]
```
JosePisco, line 14: Shouldn't be openssl in default features at this point ?

crate/kmip/src/crypto/mod.rs
ThibsG: `derive_key_from_password` that is using `Argon2` in `password_derivation.rs` is using `rand_bytes`, which is a function from openssl.
Maybe it should work without openssl, but today this is not the case

crate/kmip/Cargo.toml
ThibsG: I don't think so, Openssl comes with heavy dependencies regarding the system, and one could want to avoid those deps.

As an example, the CLI and Client crates does not need openssl, and it would be weird to require them.

crate/kmip/src/crypto/mod.rs
JosePisco: ah you are right I forgot about `rand_bytes`
I dont think it should work without it since `rand_bytes` is a secure random generator from fips approved library.

JosePisco (approved):
conversations resolved; approving

PR #190, opened 2024-02-21
Fixing inconsistent crypto consts

JosePisco:
Declared crypto consts were describing sizes in both bits and bytes resulting in inconsistent comparisons and harder to read code.

### DIFF
--- a/crate/kmip/src/crypto/password_derivation.rs
+++ b/crate/kmip/src/crypto/password_derivation.rs
@@ -14,14 +14,14 @@ use crate::kmip_bail;
 const FIPS_MIN_SALT_SIZE: usize = 16;
 #[cfg(feature = "fips")]
 /// Output size in bits of the hash function used in PBKDF2.
-const FIPS_HLEN_BITS: usize = 512;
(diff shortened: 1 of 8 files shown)


ThibsG (approved):

crate/kmip/src/crypto/password_derivation.rs
```diff
@@ -10,13 +10,16 @@ use crate::error::KmipError;
 #[cfg(feature = "fips")]
 use crate::kmip_bail;
 
+/// Minimum random salt size in bytes to use when deriving keys.
 const FIPS_MIN_SALT_SIZE: usize = 16;
```
Manuthor, line 14: `size in bytes` and `size in bits` are a bit confusing. Does it make sense to uniformize them in the same unity ? Could we adopt a naming variable convention to make it clear that those constants are in bytes or in bits. Also, I see `XXX_KLEN`, `XXX_HLEN`, `XXX_SIZE`, `XXX_SIZE_BITS`. Should we use a unique suffix to those variables (_SIZE for example?)

Manuthor (commented):
thanks for this!

PR #197, opened 2024-03-05
use closures in CC keys update

tbrezot:
This PR:
- merges the three `match` blocks into one. This leads to a slightly too big function, but it is mainly doing control flow. All actions are performed in dedicated procedures.
- master keys and user keys update procedures both grossly follow the scheme "fetch, mutate, then send".

### DIFF
--- a/crate/server/src/core/cover_crypt/rekey_keys.rs
+++ b/crate/server/src/core/cover_crypt/rekey_keys.rs
@@ -34,7 +34,7 @@ use crate::{
 pub async fn rekey_keypair_cover_crypt(
     kmip_server: &KMS,
     cover_crypt: Covercrypt,
-    msk_uid: &str,
+    msk_uid: String,
     owner: &str,
     action: RekeyEditAction,
     params: Option<&ExtraDatabaseParams>,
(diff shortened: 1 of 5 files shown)


crate/server/src/core/cover_crypt/rekey_keys.rs
```diff
@@ -31,59 +34,157 @@ use crate::{
...
+
+    Ok(ReKeyKeyPairResponse {
+        private_key_unique_identifier: UniqueIdentifier::TextString(msk_uid.to_string()),
+        public_key_unique_identifier: UniqueIdentifier::TextString(mpk_uid),
+    })
+}
+
+/// Update the master key with a new Policy
+/// (after editing the policy typically)
+pub async fn update_master_keys(
```
ackRow, line 161: The function name and docstring are no longer really accurate since this method will update the policy, master keys and even the user keys if needed

ackRow (approved):
It should be illegal to refactor my code so cleanly :100: 
I will try to run the tests locally and see why the CI is not passing

crate/server/src/core/cover_crypt/rekey_keys.rs
tbrezot: Good catch, I changed the documentation.
I kept the name since this function actually updates the master keys. Since the policy will soon be part of the MSK, I did not want to reflect it in the naming. And the USK are not updated by this function: the dedicated `update_user_secret_keys` procedure is in charge of this.

PR #199, opened 2024-03-06
Fix potential stack overflow

### LINKED ISSUE #193: Stack overflow on rekey during `integration_tests_use_ids_no_tags`
The bug is reproducible on the branch https://github.com/Cosmian/kms/tree/bug/rekey_stack_overflow by running the test `cargo test --all-features integration_tests_use_ids_no_tags`

After some debugging, the bug seems to happen at this KMIP get call: https://github.com/Cosmian/kms/blob/bug/rekey_stack_overflow/crate/server/src/core/cover_crypt/rekey_keys.rs#L199

In my other branch, I have been able to remove this call by reusing the master key from previous functions and prevent the crash.
I still created this branch to show the weird behavior because it might indicate a hidden problem.

ThibsG:
partially fixes #193

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -17,7 +17,7 @@ jobs:
       archive-name: centos7_tests
       commands: |
         cargo build --bins
-        RUST_MIN_STACK=4194304 cargo test --workspace -- --nocapture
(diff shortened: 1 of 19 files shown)


ThibsG:
superseeded by #200

PR #200, opened 2024-03-06
Reduce stack footprint

### LINKED ISSUE #193: Stack overflow on rekey during `integration_tests_use_ids_no_tags`

ThibsG:
fixes #193

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -17,7 +17,7 @@ jobs:
       archive-name: centos7_tests
       commands: |
         cargo build --bins
-        RUST_MIN_STACK=4194304 cargo test --workspace -- --nocapture
+        cargo test --workspace -- --nocapture
       artifacts: ''
 
   fips-centos7-test:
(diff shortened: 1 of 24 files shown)


JosePisco (approved):
:+1:

ackRow (approved):

tbrezot:
After a quick read of the diffs, my feelings are mixed: on the one hand, this patch sure should fix the issue; one the other hand, it adds a lot of complexity by exposing the boxing. Moreover, I struggle to see the pattern you followed to decide which object to box.

Wouldn't boxing the KMIP `Object` be enough?

ThibsG:
How course, boxing the full `Object` would do the trick, but if you use `Object` on its own or sub-components (such as `KeyBlock` or `KeyValue`) you will have the same issue again, so you have to go box deeper.

I chose to box some elements based on the size they represent in their enum/struct. The threshold is totally arbitrary: if the variable is over something like 50 bytes, I did the box. AFAIK there is no magic threshold to choose when to put on the heap

PR #209, opened 2024-03-18
Make the CLI compile on Windows and MacOS (without openssl installed)

### LINKED ISSUE #105: Removal of openssl in the CLI
This will make the CLI pure Rust and facilitate compiling to various targets.

After #101 , only the `verify.rs` need to be ported to OpenSSL.

bgrieder:
- `reqwest` is switched to `native-tls` so it should not be using openssl on Windows and Macos but will on Linux (so that it can be FIPS compliant server-side)
-  <s>`josekit` depends on openssl. Using the vendored openssl version in the client so that the `ckms` cli can build and run on Windows and MacOS without having to install openssl.</s> After discussion, JWE is nor used and its support was removed, removing the need for `josekit`
- CLI now depends on the KMS client only, which re-exposes KMIP 

Closes #105


Cargo.toml
```diff
@@ -100,5 +100,5 @@ url = "2.4"
 uuid = { version = "1.6", features = ["v4"] }
 webpki-roots = "0.22"
 x509-cert = { version = "0.2", features = ["pem"] }
-x509-parser = { version = "0.15.1", features = ["verify"] }
+x509-parser = { version = "0.16", features = ["verify"] }
```
ThibsG, line 103: can you also update `x509-parser` in `kmip` crate (dev dependencies) to unify versions? Workspace dep can be used there by the way

crate/cli/Cargo.toml
```diff
@@ -21,7 +21,8 @@ staging = []
 fips = ["cosmian_kmip/fips"]
 
 [dependencies]
-actix-web = { workspace = true }
+## Avoid pulling openssl
+actix-web = { version = "4.5.1", features = ["rustls"] }
```
ThibsG, line 25: we can keep using workspace dep here:
```suggestion
actix-web = { workspace = true, features = ["rustls"] }
```
and update the one in top-level `Cargo.toml`

ThibsG (approved):

PR #211, opened 2024-03-18
Remove KMIP dep for cli and JWE

ThibsG:
- remove KMIP dep for CLI and use client re-export
- remove JWE (and thus josekit dep that was using openssl)


JosePisco (approved):
if josekit needs removal, LGTM :+1:

PR #215, opened 2024-03-19
Run binaries in clean env also in debug mode

Manuthor:


.github/workflows/build_all_release.yml
```diff
@@ -120,11 +120,14 @@ jobs:
         sudo apt-get install --no-install-recommends -qq libclang-dev gcc-mingw-w64-x86-64
         rustup target add x86_64-pc-windows-gnu
 
+        # build only `ckms`
+        cargo build -p cosmian_kms_cli --target x86_64-pc-windows-gnu --no-default-features
```
ThibsG, line 124: no flag `--release`?

PR #254, opened 2024-06-19
fix(RUSTSEC-2024-0344): vulnerability in curve25519-dalek

Manuthor:
```
error[vulnerability]: Timing variability in `curve25519-dalek`'s `Scalar29::sub`/`Scalar52::sub`
    ┌─ /github/workspace/Cargo.lock:132:1
    │
132 │ curve25519-dalek 4.1.1 registry+https://github.com/rust-lang/crates.io-index
    │ ---------------------------------------------------------------------------- security vulnerability detected
    │
    = ID: RUSTSEC-2024-0344
    = Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0344
    = Timing variability of any kind is problematic when working with  potentially secret values such as
      elliptic curve scalars, and such issues can potentially leak private keys and other secrets. Such a
      problem was recently discovered in `curve25519-dalek`.
      
      The `Scalar29::sub` (32-bit) and `Scalar52::sub` (64-bit) functions contained usage of a mask value
      inside a loop where LLVM saw an opportunity to insert a branch instruction (`jns` on x86) to
      conditionally bypass this code section when the mask value is set to zero as can be seen in godbolt:
      
      - 32-bit (see L106): <https://godbolt.org/z/zvaWxzvqv>
      - 64-bit (see L48): <https://godbolt.org/z/PczYj7Pda>
      
      A similar problem was recently discovered in the Kyber reference implementation:
      
      <https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/hqbtIGFKIpU/m/cnE3pbueBgAJ>
      
      As discussed on that thread, one portable solution, which is also used in this PR, is to introduce a
      volatile read as an optimization barrier, which prevents the compiler from optimizing it away.
      
      The fix can be validated in godbolt here:
      
      - 32-bit: <https://godbolt.org/z/jc9j7eb8E>
      - 64-bit: <https://godbolt.org/z/x8d46Yfah>
      
      The problem was discovered and the solution independently verified by 
      Alexander Wagner <alexander.wagner@aisec.fraunhofer.de> and Lea Themint <lea.thiemt@tum.de> using
      their DATA tool:
      
      <https://github.com/Fraunhofer-AISEC/DATA>
    = Announcement: https://github.com/dalek-cryptography/curve25519-dalek/pull/659
    = Solution: Upgrade to >=4.1.3 (try `cargo update -p curve25519-dalek`)
```

- [x] Update `curve25519-dalek` from 4.1.1 to 4.1.3
- [x] must update toolchain (from january to june)
- [x] make clippy happy since new warnings are now generated with the rust-june version


crate/kmip/src/crypto/secret.rs
Adamk93, line 1: Secret is it not anymore Serializable then, right?

crate/server/src/core/operations/destroy.rs
Adamk93, line 1: I don't know why it loses the lifetime param. but I see that `ExtraDatabaseParams` does not need it anymore

crate/server/src/core/operations/revoke.rs
Adamk93, line 1: Similar to the comment done for the destroy

Adamk93 (approved):
I just added few notes, but as far as I got, to me everything seems fine. The code rewriting keeps the semantics of the programs

crate/server/src/core/operations/destroy.rs
Manuthor: the version 1.1.1 of `async-recursion` does not require anymore the explicit lifetime declaration

PR #283, opened 2024-07-24
fix: validate certificate generation

Manuthor:
- Use Validate KMIP Operation to verify certificates generated by KMS server
  * validate also self-signed certificate
- Regenerate test certificates to include a CRL linked to package.cosmian.com
- Simplify Validate flow by first ordering certificates list (then certificate chain signature and revocation status can be be verified in cascade)
- Do not verify revocation status across all chain CRLs (only parent or attached certificate CRL distribution point)
- Add small cache for CRLs (avoiding too many external requests)
- Use a workaround for the hyper `Incomplete` error by not using client certificate authentication in tests (Incomplete error being quite hard to reproduce in local - not in CI!)  
- Logs: externalize RUST_LOG variable declaration to ease tests-debugging
-  Upgrade cargo deps to last patch versions and fix `cargo deny`


crate/cli/src/main.rs
Adamk93, line 1: here you added a real log right?

Manuthor: `log_init` initializes the logger if:
- RUST_LOG variable is sourced
- or if the input argument `default_value` is filled
In that case, the CLI will have logs if RUST_LOG is sourced

Adamk93 (approved):
I have no particular comments. I put some questions, but really, curiosities. Otherwise it seems perfect to me.

crate/client/src/kms_rest_client.rs
```diff
@@ -503,12 +505,8 @@ impl KmsClient {
         // Build the client
         Ok(Self {
             client: builder
-                .connect_timeout(Duration::from_secs(10))
-                .timeout(Duration::from_secs(10))
-                .tcp_keepalive(Duration::from_secs(5))
-                .pool_idle_timeout(Duration::from_secs(5))
-                .pool_max_idle_per_host(2)
                 .default_headers(headers)
+                .tcp_keepalive(Duration::from_secs(60))
```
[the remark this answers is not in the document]
Manuthor: The problem [hyper](https://github.com/hyperium/hyper/issues/2136) is occurring (from time to time) when client authentication by certificate is enabled on the Rust Client.
I believe the 60s delay has no impact on this problem. 
Currently, I've disabled the certificate authentication wherever it is not needed (this is only a workaround). It could be fixed properly using reqwest and rustls-tls instead of native-tls.

PR #292, opened 2024-08-21
fix(validate): use certificate file path in ckms arguments

Manuthor:
From a CLI-ckms perspective, the Validate operation must use either unique ids or certificates filepath (instead of string).


Adamk93 (commented):
Nothing to say! Changed strings into path buffer and used paths to files accordingly

Adamk93 (approved):

PR #299, opened 2024-09-05
feat: build a generic database upgrade mechanism

### LINKED ISSUE #73: Build a generic database upgrade mechanism
A generic database upgrade mechanism on startup should be built.

This involves:

- Creating a `context` table with a single record holding the last version of the software run (when the table is not present, assume the version run is the one before the version implementing the change)
- On server startup:
  - the server checks if the software version is greater than the last version run
  - if no,  it simply starts
  - if yes,
    -  it looks for all upgrades to apply in order from the last version run to this version (this is done using a static function hardcoding the upgrade functions to call for each version) 
    - if there is any to run, it sets an `upgrading` flag on the `db state` field in the `context` table
    - it runs all the upgrades in order
    - it sets the flag from `upgrading` to `ready`
 - On every call to the DB, a check must be performed on the `db state` field to check if the DB is upgrading. If yes, calls should fail. Performance can be maintained using a secondary write-through cache of the `context` table

Upgrades should be programmed to resist being interrupted in the middle and resumed from start.

Manuthor:
Closes #73


crate/server/src/database/redis/redis_with_findex.rs
```diff
@@ -228,6 +228,10 @@ impl Database for RedisWithFindex {
         None
     }
 
+    async fn migrate(&self, _params: Option<&ExtraDatabaseParams>) -> KResult<()> {
+        Ok(())
+    }
```
ccorsin, line 233: maybe add a TODO, marking that it is not implemented for Redis - or better, raise an Error ?

crate/server/src/database/tests/database_tests.rs
```diff
@@ -26,7 +27,7 @@ use crate::{
 pub(crate) async fn tx_and_list<DB: Database>(
     db_and_params: &(DB, Option<ExtraDatabaseParams>),
 ) -> KResult<()> {
-    cosmian_logger::log_utils::log_init(None);
+    log_init(None);
```
ccorsin, line 30: What is this for ?

crate/server/src/database/cached_sqlcipher.rs
```diff
@@ -381,12 +442,12 @@ impl Database for CachedSqlCipher {
     async fn is_object_owned_by(
         &self,
         uid: &str,
-        userid: &str,
+        owner: &str,
```
ccorsin, line 445: keep user here? The result can be that the user is not an owner no?

Manuthor: I've renamed it to be conform to the `Database` trait:

```
    /// Test if an object identified by its `uid` is currently owned by `owner`
    async fn is_object_owned_by(
        &self,
        uid: &str,
        owner: &str,
    ) -> KResult<bool>;
```

crate/server/src/database/tests/database_tests.rs
Manuthor: it is supposed to clarify the reading ^^

crate/server/src/database/redis/redis_with_findex.rs
Manuthor: replaced by:
```
    async fn migrate(&self, _params: Option<&ExtraDatabaseParams>) -> KResult<()> {
        unimplemented!("Redis-with-Findex does not support migrate operation");
    }
```

ccorsin (approved):

PR #301, opened 2024-09-09
feat: wrap with given algo

ccorsin:


crate/client/src/export_utils.rs
```diff
@@ -85,13 +115,16 @@ fn key_wrapping_specification(
 /// * If the KMS cannot be reached
 /// * If the object cannot be exported
 /// * If the object cannot be written to a file
```
Manuthor, line 117: add the description of the 2 new params

crate/client/src/export_utils.rs
```diff
@@ -85,13 +115,16 @@ fn key_wrapping_specification(
...
+#[allow(clippy::too_many_arguments)]
```
Manuthor, line 118: create also a struct here and remove the clippy lint

crate/kmip/src/crypto/wrap/wrap_key.rs
```diff
@@ -120,6 +140,7 @@ pub(crate) fn wrap(
     wrapping_key: &Object,
     key_wrapping_data: &KeyWrappingData,
     key_to_wrap: &[u8],
+    additionnal_data_encryption: &[u8],
```
Manuthor, line 143:
```suggestion
    additional_data_encryption: Option<&[u8]>,
```
make it optional?

crate/kmip/src/kmip/kmip_types.rs
```diff
@@ -1162,6 +1163,29 @@ impl Attributes {
 
         Ok((usage_mask & flag).bits() != 0)
     }
+
```
Manuthor, line 1166: finally, is it required to use vendor_attribute?

crate/cli/src/actions/shared/export_key.rs
```diff
@@ -24,6 +25,12 @@ pub enum ExportKeyFormat {
     Raw,
 }
 
+#[derive(clap::ValueEnum, Debug, Clone, PartialEq, Eq)]
+pub(crate) enum ExportBlockCipherMode {
```
Manuthor, line 29: I wonder if we could use instead directly the KMIP BlockCipherMode enum (and add the nist-key-wrap as extension to it). Avoiding us to declare and maintain a new enum.

ccorsin: I tried first - but wasn't able to access it from the CLI's folder; maybe we can watch this together next week

PR #308, opened 2024-09-16
Fix/windows build

Manuthor:
- adapt Docker images (FIPS and not FIPS)
- adapt Debian and RPM packages
- fix test `Clean env. Windows`


bgrieder (approved):
LGTM

PR #314, opened 2024-09-20
test: wrap export import

Manuthor:


crate/kmip/src/crypto/wrap/wrap_key.rs
```diff
@@ -187,17 +218,15 @@ pub(crate) fn wrap(
...
                         Some(BlockCipherMode::GCM) => {
                             // wrap using aes Gcm
                             let aead = AeadCipher::Aes256Gcm;
                             let nonce = random_nonce(aead)?;
                             let (data, authenticated_encryption_tag) =
                                 aead_encrypt(aead, &wrap_secret, &nonce, aad, key_to_wrap)?;
-                            let ciphertext =
-                                [nonce.clone(), data, authenticated_encryption_tag.clone()]
-                                    .concat();
+                            let ciphertext = [nonce, data, authenticated_encryption_tag].concat();
```
ccorsin, line 229: didn't you wanted to remove concat and use extend_from_slice function ?

crate/cli/src/tests/shared/import.rs
```diff
@@ -26,6 +26,7 @@ use crate::{
 
 #[derive(Default, Debug)]
 pub(crate) struct ImportKeyParams {
+    // TODO: should be replaced by ImportKeyAction
```
ccorsin, line 29: Should we do here, or in the merging branche ?

crate/kmip/src/crypto/wrap/wrap_key.rs
Manuthor: exact!

crate/cli/src/tests/shared/import.rs
Manuthor: maybe in a another PR if you are OK (because it is not a transparent change, we need to change the function `import_key` signature a little to reuse the `ImportKeyAction` struct. The same for `export_key`.)

ccorsin: ok !

PR #327, opened 2024-10-08
fix: replace Debug derive trait of KMIP Object by a custom Display impl

Manuthor:
- make sure no key content can be displayed for logging purpose by replacing the `Debug` derive trait of KMIP `Object` (and other KMIP types) by a custom Display implementation
- replace log formatting `{:?}` with Display impl `{}`
- something seems to be wrong in `batch_operations` function: simplifying return type from `Result<Vec<Result<Operation, String>>, ClientError>` to `Result<Vec<Operation>, ClientError>`

### DIFF
--- a/crate/cli/src/actions/cover_crypt/decrypt.rs
+++ b/crate/cli/src/actions/cover_crypt/decrypt.rs
@@ -87,7 +87,7 @@ impl DecryptAction {
             }),
         );
 
-        tracing::debug!("{decrypt_request:?}");
+        tracing::debug!("{decrypt_request}");
 
         // Query the KMS with your kmip data and get the key pair ids
         let decrypt_response = kms_rest_client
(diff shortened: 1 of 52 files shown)


Manuthor:
> Cool pour les `Display` Pas de pb avec le fix du `Result` inutile sur les batchs Pourquoi as-tu changé des `assert_eq!(a,b)` en `assert!(a==b)` ?

j'ai ce type d'erreur sinon:
```
error[E0277]: `cosmian_kmip::kmip::kmip_objects::Object` doesn't implement `std::fmt::Debug`
  --> crate/server/src/database/tests/tagging_tests.rs:85:5
   |
85 |     assert_eq!(symmetric_key, owm.object);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `cosmian_kmip::kmip::kmip_objects::Object` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
   |
   = help: the trait `std::fmt::Debug` is not implemented for `cosmian_kmip::kmip::kmip_objects::Object`
   = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)

   Compiling kms_test_server v4.19.0 (/home/manu/Cosmian/core/kms_alt/crate/test_server)
```

PR #341, opened 2024-11-22
feat: make keys non revokable

Manuthor:
- Server: 
  - add a clap option with `non revokable key ids`
  - on revoke operation, dismiss the revoke operation on key id found in those `non revokable key ids` list

- CLI:
  - add test to check revocation  
  -  on create symmetric key, change behavior and add key_id as argument (seems to be more consistent with other commands)
  - reuse CreateKeyAction on tests (to simplify test functions arguments and to benefit of the key_id on sym key creation)

### DIFF
--- a/crate/server/src/config/command_line/clap_config.rs
+++ b/crate/server/src/config/command_line/clap_config.rs
@@ -21,6 +21,7 @@ impl Default for ClapConfig {
             ms_dke_service_url: None,
             telemetry: TelemetryConfig::default(),
             info: false,
+            non_revokable_key_id: None,
         }
     }
 }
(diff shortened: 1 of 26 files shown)


crate/server/src/config/command_line/clap_config.rs
```diff
@@ -21,6 +21,7 @@ impl Default for ClapConfig {
             ms_dke_service_url: None,
             telemetry: TelemetryConfig::default(),
             info: false,
+            non_revokable_key_id: None,
```
ccorsin, line 24: make it plural ?

documentation/docs/cli/main_commands.md
```diff
@@ -1769,7 +1768,7 @@ Possible values:  `"chacha20", "aes", "sha3", "shake"` [default: `"aes"`]
 
 `--tag [-t] <TAG>` The tag to associate with the key. To specify multiple tags, use the option multiple times
 
-` <KEY_ID>` The unique id of the key; a random uuid is generated if not specified
+`--key-id <KEY_ID>` The unique id of the key; a random uuid is generated if not specified
```
ccorsin, line 1771: To be consistant, maybe handle this option the same way for other KMIP creation objects ?

ccorsin (approved):

PR #376, opened 2025-02-27
Draft: Wasm

ccorsin:


crate/ui-utils/src/certificate_utils.rs
Manuthor, line 1: we can remove the `_utils` suffix in all the files

PR #385, opened 2025-03-19
Review of the Covercrypt integration

tbrezot:
This MR:
- drops the use of the encrypted header, hence removing the extraneous metadata argument up to the very KMS API
- removes the policy from the attributes of the master keys;
- makes use of a now stable feature: async closures. I chose to upgrade the nightly toolchain instead of activating yet another compilation flag and benefit from the last modification of this recent feature;
- removes lots of unnecessary clones by taking ownership everywhere it is needed. Some clones remain but the data flow is much cleaner (imho);
- renames a bunch of variables/procedures to use common short names (USK, MSK, MPK) : `update_usk` is arguably easier to read than `update_user_decryption_key`, while containing the same information;
- renames a bunch of variables which name is redundant with their type/context (e.g. `let req = build_create_covercrypt_master_keypair_request` instead of `let create_request = ...` which is shorter but does not loose information since the context already makes it clear).
- removes comment that are redundant with the code, e.g.: 
```        
        // only private keys
        if owm.object().object_type() != ObjectType::PrivateKey {
            continue
        }
```

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -41,6 +41,11 @@ jobs:
         run: cargo +${{ inputs.toolchain }} fmt --all -- --check --color always
 
       - name: Static analysis
+        run: cargo clippy --workspace --all-targets -- -D warnings
+        env:
+          OPENSSL_DIR: ${{ env.OPENSSL_DIR }}
+
+      - name: Static analysis all features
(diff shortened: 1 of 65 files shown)


Manuthor (approved):
Great simplification!

PR #447, opened 2025-06-12
Feat/register operation

ccorsin:
On this PR :
- Add Register KMIP operation (server side only)
- Handle KMIP Date-Time as OffsetDateTime instead of i64
- Test KMIP Register and Import on KMIP 2.1 & 1.4

### DIFF
--- a/crate/cli/Cargo.toml
+++ b/crate/cli/Cargo.toml
@@ -58,7 +58,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 strum = { workspace = true }
 test_kms_server = { path = "../test_kms_server", version = "5.4.1" }
-time = "0.3"
+time = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 tracing = { workspace = true }
--- a/crate/client_utils/Cargo.toml
+++ b/crate/client_utils/Cargo.toml
@@ -27,7 +27,7 @@ pem = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 strum = { workspace = true, features = ["std", "derive", "strum_macros"] }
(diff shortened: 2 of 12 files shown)


crate/cli/Cargo.toml
```diff
@@ -58,6 +58,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 strum = { workspace = true }
 test_kms_server = { path = "../test_kms_server", version = "5.4.1" }
+time = "0.3"
```
bgrieder: time should be registered in the workspace toml and a reference `{workspace = true} ` maintained here

crate/client_utils/Cargo.toml
```diff
@@ -27,6 +27,7 @@ pem = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 strum = { workspace = true, features = ["std", "derive", "strum_macros"] }
+time = "0.3"
```
bgrieder: see above

crate/server/src/core/operations/register.rs
```diff
@@ -0,0 +1,106 @@
...
+    if request.protection_storage_masks.is_some() {
+        kms_bail!(KmsError::UnsupportedPlaceholder)
+    }
+
+    // To register an object, check that the user has `Create` access right
+    if let Some(users) = privileged_users.clone() {
+        let has_permission = user_has_permission(
+            owner,
+            None,
+            &cosmian_kmip::kmip_2_1::KmipOperation::Create,
```
bgrieder, line 47: We need to comment here that the delegation of Create covers Create Import and Register
The documentation must also be clear on that aspect

bgrieder (commented):
A few minor changes

bgrieder (approved):
LGTM

PR #473, opened 2025-07-16
adding_percona_doc

chokoblitz:
Here is the documentation for installing the Cosmian Kms with postgresql percona


bgrieder (approved):
LGTM

PR #493, opened 2025-08-01
feat: support AES-CBC encryption without padding

Manuthor:
- In case of Oracle TDE integration, AES-CBC encryption without padding is needed. PKCS7 padding (formerly PKCS5) is kept by default for CBC mode
- install last CA roots on docker image
- also add import of `OpaqueObject`

### DIFF
--- a/crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
+++ b/crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
@@ -359,14 +359,14 @@ pub fn encrypt(
             let padding = padding_method.unwrap_or(PaddingMethod::PKCS5);
             let ciphertext = match padding {
                 PaddingMethod::None => {
-                    if plaintext.len() % 16 != 0 {
+                    let cipher = sym_cipher.to_openssl_cipher()?;
+                    if plaintext.len() % cipher.block_size() != 0 {
                         return Err(CryptoError::InvalidSize(
                             "Plaintext must be a multiple of the block size when no padding is \
                              used"
                                 .to_owned(),
                         ));
                     }
-                    let cipher = sym_cipher.to_openssl_cipher()?;
                     let mut c =
                         Crypter::new(cipher, openssl::symm::Mode::Encrypt, key, Some(nonce))?;
                     c.pad(false);
@@ -450,14 +450,14 @@ pub fn decrypt(
                     ciphertext,
                 )?),
                 PaddingMethod::None => {
-                    if ciphertext.len() % 16 != 0 {
+                    let cipher = sym_cipher.to_openssl_cipher()?;
+                    if ciphertext.len() % cipher.block_size() != 0 {
                         return Err(CryptoError::InvalidSize(
                             "Ciphertext must be a multiple of the block size when no padding is \
                              used"
                                 .to_owned(),
                         ));
                     }
-                    let cipher = sym_cipher.to_openssl_cipher()?;
                     let mut c =
                         Crypter::new(cipher, openssl::symm::Mode::Decrypt, key, Some(nonce))?;
                     c.pad(false);
@@ -473,7 +473,7 @@ pub fn decrypt(
                 }
                 _ => {
                     return Err(CryptoError::NotSupported(format!(
-                        "Padding method {padding_method:?} is not currently supported"
+                        "Padding method {padding:?} is not currently supported"
                     )));
                 }
             }
--- a/crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
+++ b/crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
@@ -245,38 +245,42 @@ impl SymCipher {
                     BlockCipherMode::AEAD | BlockCipherMode::GCM => match key_size {
                         AES_128_GCM_KEY_LENGTH => Ok(Self::Aes128Gcm),
                         AES_256_GCM_KEY_LENGTH => Ok(Self::Aes256Gcm),
-                        _ => crypto_bail!(CryptoError::NotSupported(
-                            "AES key must be 16 or 32 bytes long for AES GCM".to_owned()
-                        )),
+                        _ => crypto_bail!(CryptoError::NotSupported(format!(
+                            "AES key must be 16 or 32 bytes long for AES GCM. Found {key_size} \
+                             bytes",
+                        ))),
                     },
                     BlockCipherMode::CBC => match key_size {
                         AES_128_CBC_KEY_LENGTH => Ok(Self::Aes128Cbc),
                         AES_256_CBC_KEY_LENGTH => Ok(Self::Aes256Cbc),
-                        _ => crypto_bail!(CryptoError::NotSupported(
-                            "AES key must be 16 or 32 bytes long for AES CBC".to_owned()
-                        )),
+                        _ => crypto_bail!(CryptoError::NotSupported(format!(
+                            "AES key must be 16 or 32 bytes long for AES CBC. Found {key_size} \
+                             bytes",
+                        ))),
                     },
                     BlockCipherMode::XTS => match key_size {
                         AES_128_XTS_KEY_LENGTH => Ok(Self::Aes128Xts),
                         AES_256_XTS_KEY_LENGTH => Ok(Self::Aes256Xts),
-                        _ => crypto_bail!(CryptoError::NotSupported(
-                            "AES key must be 32 or 64 bytes long for AES XTS".to_owned()
-                        )),
+                        _ => crypto_bail!(CryptoError::NotSupported(format!(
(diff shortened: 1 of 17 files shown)


crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
```diff
@@ -245,38 +246,42 @@ impl SymCipher {
                     BlockCipherMode::AEAD | BlockCipherMode::GCM => match key_size {
                         AES_128_GCM_KEY_LENGTH => Ok(Self::Aes128Gcm),
                         AES_256_GCM_KEY_LENGTH => Ok(Self::Aes256Gcm),
-                        _ => crypto_bail!(CryptoError::NotSupported(
-                            "AES key must be 16 or 32 bytes long for AES GCM".to_owned()
-                        )),
+                        _ => crypto_bail!(CryptoError::NotSupported(format!(
```
HatemMn, line 249: _(just for learning puposes)_

I don't understand the presence of this macro, like
- what does crypto bail mean ( googling it does not help )
- what's the difference with the other errors, ie 
```rust
return Err(CryptoError::InvalidSize(
                            "Plaintext must be a multiple of the block size when no padding is \
                             used"
                                .to_owned(),
                        ));
```

crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
```diff
@@ -352,6 +355,42 @@ pub fn encrypt(
...
+                    let cipher = sym_cipher.to_openssl_cipher()?;
+                    if plaintext.len() % cipher.block_size() != 0 {
+                        return Err(CryptoError::InvalidSize(
+                            "Plaintext must be a multiple of the block size when no padding is \
+                             used"
+                                .to_owned(),
+                        ));
+                    }
+                    let mut c =
+                        Crypter::new(cipher, openssl::symm::Mode::Encrypt, key, Some(nonce))?;
```
HatemMn, line 371: _(just for learning puposes)_

Is it cummon in this codebase to call the crypters "c" or it's just a local choice here ?

crate/crypto/src/crypto/symmetric/tests.rs
```diff
@@ -100,12 +128,84 @@ fn test_encrypt_decrypt_aes_xts_256() {
...
+        &ciphertext,
+        &tag,
+        None,
+    )
+    .unwrap();
+
+    // `to_vec()` conversion because of Zeroizing<>.
+    assert_eq!(decrypted_data.to_vec(), message);
+}
+
```
HatemMn, line 177: is a test needed for the last case ? ( ie : a test that gives some other padding methods, and the result is getting an eeror for non supported padding methods )

crate/crypto/src/crypto/symmetric/tests.rs
```diff
@@ -100,12 +128,84 @@ fn test_encrypt_decrypt_aes_xts_256() {
...
+    let decrypted_data = decrypt(
+        SymCipher::Aes256Cbc,
+        &key,
+        &tweak,
+        &[],
+        &ciphertext,
+        &tag,
+        None,
+    )
+    .unwrap();
```
HatemMn:
```suggestion
    let (ciphertext, tag) = encrypt(cipher, &key, &tweak, &[], &message, Some(PaddingMethod::PKCS5)).unwrap();

    let decrypted_data = decrypt(
        SymCipher::Aes256Cbc,
        &key,
        &tweak,
        &[],
        &ciphertext,
        &tag,
        Some(PaddingMethod::PKCS5),
    )
    .unwrap();
```

From what I understand : this is the unit test responsible of testing the PaddingMethod PKCS5, however, no padding is used as opposed to the actual name of the test

crate/crypto/src/crypto/wrap/unwrap_key.rs
```diff
@@ -272,7 +272,15 @@ fn aes_gcm_unwrap(
...
         .ok_or_else(|| CryptoError::IndexingSlicing("unwrap: tag".to_owned()))?;
-    decrypt(aead, unwrap_secret, nonce, &[], wrapped_key_bytes, tag)
+    decrypt(
+        aead,
+        unwrap_secret,
+        nonce,
+        &[],
+        wrapped_key_bytes,
+        tag,
+        None,
```
HatemMn, line 282: I don't understand a lot how wrapping works but I have a question, reading the diffs in `wrap_key.rs` I see that `Some(padding_method)` is provided as argument, but here however it's always set to `None`. Is that fine ?

crate/server/src/core/operations/import.rs
```diff
@@ -837,3 +838,41 @@ pub(crate) async fn process_secret_data(
         )],
     ))
 }
+
+pub(crate) fn process_opaque_object(
```
HatemMn, line 842: _(self documentation)_

what's an opaque object, and what does this file do ?

crate/crypto/src/crypto/symmetric/tests.rs
```diff
@@ -100,12 +128,84 @@ fn test_encrypt_decrypt_aes_xts_256() {
...
+
+    // `to_vec()` conversion because of Zeroizing<>.
```
HatemMn: sorry I didn't get this comment

HatemMn (commented):
LGTM ! Most of my questions are simply for my own docs, I indicated when it's the case

By the way, does the crate need a bump or that's automatically done by some CD system ?

crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
Manuthor: `crypto_bail` slightly shortens the syntax for raising an error.

crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
Manuthor: It is a local choice only

crate/crypto/src/crypto/symmetric/tests.rs
Manuthor: adding a test with unsupported PaddingMethod::ANSI_X923

crate/crypto/src/crypto/symmetric/tests.rs
Manuthor: Yes because by default, AES-CBC uses the PKCS5 padding. But you're right, let's mix up: encrypt without padding specified and decrypt with PKCS5 padding (explicitly specified)

crate/crypto/src/crypto/symmetric/tests.rs
Manuthor: Remove them.
(The .to_vec() is used next line to compare 2 Vec<u8>)

crate/server/src/core/operations/import.rs
Manuthor: The initial approach I've chosen to resolve the Oracle TDE PKCS11 calls was to use Opaque Object to store object that has no cryptographic specificities. 
I've kept this code because it does not harm anything (one could need one day to import an opaque object)

crate/crypto/src/crypto/wrap/unwrap_key.rs
Manuthor: Yes it is fine. This `decrypt` generic function does not have always a supported padding method.

crate/crypto/src/crypto/symmetric/tests.rs
HatemMn: I think we can do them all ? It might help with code coverage scores (if there are some, here)

```rust
    /**
         *
    /// Padding Method Enumeration
    pub enum PaddingMethod {
        None = 0x1,
        OAEP = 0x2,
        PKCS5 = 0x3,
        SSL3 = 0x4,
        Zeros = 0x5,
        // #[serde(rename = "ANSI X9.23")]
        ANSI_X923 = 0x6,
        // #[serde(rename = "ISO 10126")]
        ISO10126 = 0x7,
        // #[serde(rename = "PKCS1 v1.5")]
        PKCS1v15 = 0x8,
        // #[serde(rename = "X9.31")]
        X931 = 0x9,
        PSS = 0xA,
    }
         */
    let padding_options = [OAEP, SSL3, Zeros, ANSI_X923, ISO10126, PKCS1v15, X931, PSS];
    let mut random_byte = [0_u8; 1];
    rand_bytes(&mut random_byte).unwrap();
    let index = random_byte[0] as usize % padding_options.len();
    let random_padding_method = padding_options[index];
```

Manuthor: done in debe5cb425dfd4d8ad1c51e56a485fc67c45841f

Manuthor:
> LGTM ! Most of my questions are simply for my own docs, I indicated when it's the case
> 
> By the way, does the crate need a bump or that's automatically done by some CD system ?

I'll do the bump manually when releasing.

PR #516, opened 2025-08-27
fix: keep error info on DBerror

HatemMn:
_(this is a preparatory PR for https://github.com/Cosmian/kms/pull/542 in order to divide concerns )_

## Overview
A lot of errors conversion is made using `to_string` in the following file `crate/server_database/src/error/db_error.rs`

However, [the thiserror crate](https://docs.rs/thiserror/latest/thiserror/) is already a part of the project, and not using it is rather a sort of bottleneck since : 
- We manually rewrite code that the library can abstract using its macros
- We convert errors to strings, losing a lot of context in the process

⚠️ This update won't recode all the error handling of the KMS, it's mostly a POC of the fact that we **can** handle errors in a better way. If this PR gets merged, we might consider generalising its approach to all the codebase - which will drastically reduce boilerplate error code and make recoverable errors more ... recoverable.

Also, I do need it for https://github.com/Cosmian/kms/pull/514 because the new errors that will be introduced into the enum ( **RedisMemoryError** and **FindexError** don't implement Clone - like most errors, actually )

## Why this can be useful : 

Take this example of an user handling DBerrors (current approach) : 
```rust
fn handle_database_issue(err: &DbError) -> SomeRecoveryAction {
    if let DbError::DatabaseError(msg) = err {
        // USER IS STUCK! They only have the string message 🫨 
        if msg.contains("pool") && msg.contains("timed out") {
            // Fragile string parsing - not fun 🚫 
            RecoveryAction::IncreasePoolSize
        }
```

With the new code : 

```rust
// A user debugging the KMS can do this instead:
fn handle_database_issue(err: &DbError) -> SomeRecoveryAction {
    if let DbError::DatabaseError(sqlx_err) = err {
        match sqlx_err {
            sqlx::Error::PoolTimedOut => {
                // easy and infallible
                log::warn!("Connection pool exhausted - scaling up pool size");
                RecoveryAction::IncreasePoolSize
            }
            sqlx::Error::Database(db_err) if db_err.code() == Some("23505") => {
                // (another example) user can detect PostgreSQL unique constraint violation
                RecoveryAction::RetryWithNewKey
            }
           // etc, more err handling
```

## ⚠️ Why I had to update the UnwrappedCache

This wasn't implemented before because the error type needed to be clonable, but usually library errors don't bother to derive `Clone`

I had to change the cache error handling a bit in order to keep an Arc<...> to the DBError. ( this changes nothing of the functionning of the codebase). Due to Rust limitations, the cache needs a clonable error and Arc<Error> does the trick )

This way, instead of having a full String errors because of an obnoxious Cache object, we have the cache itself return a string error, and not the rest


Manuthor (approved):

crate/server/src/error.rs
```diff
@@ -40,7 +40,7 @@ pub enum KmsError {
 
     // Any errors related to a bad behavior of the DB but not related to the user input
     #[error("Database Error: {0}")]
-    Database(String),
+    Database(#[from] DbError),
```
HatemMn, line 43: using this error crate for auto conversions

crate/server/src/socket_server.rs
```diff
@@ -196,12 +196,13 @@ impl SocketServer {
...
             Err(e) => {
+                let error_msg = e.to_string(); // keep the message before moving the error
                 if let Some(notifier) = start_notifier {
                     notifier
-                        .send(Err(e.clone()))
+                        .send(Err(e))
                         .context("Failed to notify the error on socket server start")?;
                 }
-                kms_bail!("Failed to bind to {addr}: {}", e);
+                kms_bail!("Failed to bind to {addr}: {}", error_msg);
```
HatemMn, line 204: @bgrieder if you can please confirm this is "ok" before we merge, as it's the only "change of logic" of this update

all the rest changes nothing of how the kms handles errors, and rather makes it more flexible

Manuthor: LGTM

PR #525, opened 2025-09-01
fix: uniformize clippy lints on all crates

Manuthor:


crate/cli/src/actions/kms/symmetric/decrypt.rs
```diff
@@ -211,7 +211,7 @@ impl DecryptAction {
         decrypt_response.data.context("the plain text is empty")
     }
 
-    #[expect(clippy::too_many_arguments)]
+    #[expect(clippy::too_many_arguments, clippy::indexing_slicing)]
```
bgrieder, line 214: Can't see where this indexing_slicing comes from

crate/hsm/base_hsm/src/hsm_lib.rs
```diff
@@ -161,6 +161,6 @@ impl HsmLib {
 
 impl Drop for HsmLib {
     fn drop(&mut self) {
-        let _ = self.finalize();
+        drop(self.finalize());
```
bgrieder, line 171: I am surprised that clippy prefers the new version

crate/hsm/base_hsm/src/session/session_impl.rs
```diff
@@ -1211,6 +1211,6 @@ impl Session {
 
 impl Drop for Session {
     fn drop(&mut self) {
-        let _ = self.close();
+        drop(self.close());
```
bgrieder, line 1223: same here

crate/server/src/main.rs
```diff
@@ -25,11 +25,13 @@ mod expiry;
 async fn main() -> KResult<()> {
     // Set up environment variables and logging options
     if std::env::var("RUST_BACKTRACE").is_err() {
+        #[allow(unsafe_code)]
```
bgrieder: It is too bad that we have these unsafe blocks in main. We should really try to find some other solution, such as setting a singleton conf variable

crate/hsm/base_hsm/src/hsm_lib.rs
Manuthor: This is the error:
```
error: non-binding let on a type that has a destructor
   --> crate/hsm/base_hsm/src/hsm_lib.rs:164:9
    |
164 |         let _ = self.finalize();
    |         ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D let-underscore-drop` implied by `-D let-underscore`
    = help: to override `-D let-underscore` add `#[allow(let_underscore_drop)]`
help: consider binding to an unused variable to avoid immediately dropping the value
    |
164 |         let _unused = self.finalize();
    |              ++++++
help: consider immediately dropping the value
    |
164 -         let _ = self.finalize();
164 +         drop(self.finalize());
    |

```

Manuthor: I didn't fix all clippy lints for hsm crates but I'd like to uncomment this part on those Cargo.toml:
```
# [lints]
# workspace = true
```

Manuthor:
> Ahh... the `to_string()` --> `to_owned()`

I know, sorry for this. 
Generally I use clippy for this mass renaming task:
```
cargo clippy --all-targets --all-features --workspace --fix -Z unstable-options --allow-dirty --allow-staged --no-deps -- --no-deps
```

crate/server/src/main.rs
Manuthor: You are right.
Actually we can remove all the setting of environment variables (RUST_BACKTRACE and RUST_LOG) since it is  already done after in function `cosmian_logger->tracing_init` https://github.com/Cosmian/kms/blob/22e0e76cd9238f832bd916f7acb8a3fc0e8140db/crate/server/src/main.rs#L55

crate/cli/src/actions/kms/symmetric/decrypt.rs
Manuthor: It comes from here:
```
        loop {
            let bytes_read = input_file.read(&mut chunk)?;
            if bytes_read == 0 {
                break;
            }
            chunk.truncate(bytes_read);
            let available_bytes = [read_buffer.as_slice(), &chunk].concat();
            // keep at least the tag size in the local buffer
            if available_bytes.len() > tag_size {
                // process all bytes except the tag length last bytes
                let num_bytes_to_process = available_bytes.len() - tag_size;
                **let output = stream_cipher.update(&available_bytes[..num_bytes_to_process])?;**
                output_file.write_all(&output)?;
                // keep the remaining bytes in the read buffer
                **read_buffer = available_bytes[num_bytes_to_process..].to_vec();**
            } else {
                // put everything in the read buffer
                read_buffer = available_bytes;
            }
        }
```

PR #526, opened 2025-09-01
fix: use crypto core for db crypto 

HatemMn:
This is a part of cloudproof -> findex v8 update

Basically, whenever using rand_rng and and Secrets, we need to call crypto_core (not "kms_crypto") otherwise we get type conflicts with the (new) crate `cosmian_sse_memories` and findex version >= 7.0

This pr also reduces reexporting boilerplate, read the comments below to understand more

Ah and yes, I bumped `cosmian_crypto_core` to 10.2

### DIFF
--- a/crate/server_database/Cargo.toml
+++ b/crate/server_database/Cargo.toml
@@ -22,10 +22,10 @@ non-fips = [
 [dependencies]
 async-trait = { workspace = true }
 cosmian_kms_crypto = { path = "../crypto", version = "5.7.1" }
-cloudproof_findex = { version = "5.0", optional = true }
 cosmian_crypto_core = { workspace = true }
 cosmian_kmip = { path = "../kmip", version = "5.7.1" }
 cosmian_kms_interfaces = { path = "../interfaces", version = "5.7.1" }
+cloudproof_findex = { version = "5.0", optional = true }
 hex = { workspace = true }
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -6,7 +6,6 @@
     "chacha",
     "ciphertext",
     "ciphertexts",
-    "cosmian",
```
HatemMn, line 9: editing this file is indeed off topic but I did it anyway to get less CSpell spam

crate/crypto/src/lib.rs
```diff
@@ -66,9 +66,9 @@ pub fn pad_be_bytes(bytes: &mut Vec<u8>, size: usize) {
     }
 }
 
+// TODO(review comment): reexporting is not useful, version should be managed within the workspace
 pub mod reexport {
     #[cfg(feature = "non-fips")]
     pub use cosmian_cover_crypt;
-    pub use cosmian_crypto_core;
```
HatemMn, line 72: cosmian_crypto_core is defined in the worskpace root ( even before this PR existed )

this makes this reexport obsolete
either : 
- Define on root, no reexports
- Defin at crate/crypto level, keep reexports

I chosed the first path to reduce boilerplate code bloat

Manuthor (approved):
Good!

PR #536, opened 2025-09-07
feat(tracing): use comian_logger to logs instead of tracing

Manuthor:
Add automatically function names when using tracing macros

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1217,7 +1217,6 @@ dependencies = [
  "thiserror 2.0.16",
  "time",
  "tokio",
- "tracing",
  "url",
  "uuid",
  "x509-cert",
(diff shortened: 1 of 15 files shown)


crate/crypto/src/crypto/cover_crypt/encryption.rs
```diff
@@ -141,7 +141,7 @@ impl EncryptionSystem for CoverCryptEncryption {
     fn encrypt(&self, request: &Encrypt) -> Result<EncryptResponse, CryptoError> {
         let ad = request.authenticated_encryption_additional_data.as_deref();
 
-        trace!("CoverCryptEncryption: encrypt: authenticated_encryption_additional_data: {ad:?}",);
+        trace!("encryption_policy: {ad:?}",);
```
Manuthor:
```suggestion
        trace!("authenticated_encryption_additional_data: {ad:?}",);
```

PR #552, opened 2025-09-17
Add HSM key search with basic filters

SG-O:
This PR introduces support for searching for HSM objects with basic filtering options.

Keys can be filtered by:

Owner (HSM Admin)
Unique Identifier
Cryptographic Algorithm
Object Type
Cryptographic Length

As many attributes are not available for Keys stored on a HSM, I added rules to return no results when searching for certain attributes. Determining which cases to handle how wasn't easy, and should changes be need, I'm happy to make those.

Changelog:
### 🚀 Features
- cosmian_kms_interfaces: Added basic search for objects/keys stored on HSMs
- cosmian_kms_interfaces: Implemented filtering for attributed applicable to HSM stored keys.
- cosmian_kms_base_hsm: Added a function to retrieve a list of slot IDs an HSM has been initialized with.

### 🧪 Testing
- softhsm2_pkcs11_loader: Added tests for the `get_available_slot_list` function in `cosmian_kms_base_hsm`.
- cosmian_kms_server: Added tests for the search of keys stored on HSMs implemented in `cosmian_kms_interfaces`.


SG-O:
That is no problem. Should I close this PR, and reopen it to sign?
I also have two additional questions:
Should I also rework this PR to follow the new requirements outlined in CONTRIBUTING (especially the addition of a Changelog)?
Should / may I push these changes to a new branch on this repo (Cosmian:develop_hsm_find) to enable CI?

SG-O:
I added a changelog entry (please let me know if the format matches your expectations. I couldn’t find any examples so I interpreted the description)

I merged the upstream changes and most CI tests are now passing. However it tries to push to `package.cosmian.com` which it is stuck at: https://github.com/SG-O/cosmian-kms/actions/runs/17828819845/job/50690289526.

Regarding the CLA signing: what’s the preferred process? Should I open a dummy PR to interact with the bot, or is there another way?

Manuthor:
About changelog, we keep everything in the file ./CHANGELOG.md

SG-O:
> About changelog, we keep everything in the file ./CHANGELOG.md

Ok, I was complying [with this rule in the new CONTRIBUTING documentation.](https://github.com/Cosmian/kms/blob/develop/CONTRIBUTING.md#changelog-entries)

I'll get rid of the changelog I created and add its content to the PR for reference.

SG-O:
I have read the CLA Document and I hereby sign the CLA

Manuthor:
> > My bad. I've just moved the CLA assistant workflow to the right place...
> 
> Don't worry about it 😄 Signing seems to work, but it didn't remember that I already signed the CLA [in another PR](https://github.com/Cosmian/kms/pull/561).

Thank you for your patience. Not very familiar with this GitHub Action `contributor-assistant/github-action`. I've tried to simplify the main_base.yml workflow and also added the secret PERSONAL_ACCESS_TOKEN to repo, it that could be responsible of writing the signature to the upstream repo with CLA is signed.
I will finally read your PR as soon as possible.

crate/hsm/base_hsm/src/base_hsm.rs
```diff
@@ -112,6 +112,36 @@ impl<P: HsmProvider> BaseHsm<P> {
         }
     }
 
+    /// Retrieve the list of slot identifiers for the HSM.
+    ///
+    /// This function returns the IDs of all slots that the HSM has been
+    /// initialized with, typically from configuration.
+    ///
+    /// # Returns
+    /// * `HResult<Vec<usize>>` - A result containing a vector of slot identifiers
```
Manuthor:
```suggestion
    /// * A result containing the available slot identifiers
```

crate/interfaces/src/hsm/hsm_store.rs
```diff
@@ -236,9 +237,178 @@ impl ObjectsStore for HsmStore {
         user_must_be_owner: bool,
         params: Option<Arc<dyn SessionParams>>,
     ) -> InterfaceResult<Vec<(String, State, Attributes)>> {
-        // HSM Objects cannot be searched
-        Ok(vec![])
+        let slot_ids = self.hsm.get_available_slot_list().await?;
+        let mut uids = Vec::new();
+        if user_must_be_owner && user != self.hsm_admin {
+            return Ok(uids);
```
Manuthor, line 247: I would raise an error here

crate/interfaces/src/hsm/hsm_store.rs
```diff
@@ -236,9 +237,178 @@ impl ObjectsStore for HsmStore {
...
+        }
+
+        let search_attributes =
+            researched_attributes.map_or_else(Attributes::default, |search_attributes| {
+                debug!("Search attributes: {:?}", search_attributes);
+                search_attributes.clone()
+            });
```
Manuthor:
```suggestion
        let Some(search_attributes) = researched_attributes else {
                return Err(InterfaceError::InvalidRequest(
                    "HSM find requests must have attributes specified".to_owned(),
                ))
        };
        debug!("Search attributes: {search_attributes}");
```

crate/kmip/src/kmip_2_1/kmip_attributes.rs
```diff
@@ -47,7 +47,7 @@ use crate::{
 /// the server MAY retain all, some or none of the object attributes,
 /// depending on the object type and server policy.
 // TODO: there are 56 attributes in the specs. Only a handful are implemented here
-#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Default)]
+#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Default, Debug)]
```
Manuthor:
```suggestion
#[derive(Serialize, Deserialize, Clone, Eq, PartialEq, Default)]
```

Manuthor: I would rather not using the Debug impl since we have now the Display impl that allow a compact print.

crate/interfaces/src/hsm/interface.rs
```diff
@@ -90,6 +91,16 @@ impl HsmObject {
...
 #[async_trait]
 pub trait HSM: Send + Sync {
+    /// Get the list of available slot identifiers for the HSM.
+    ///
+    /// This function retrieves the identifiers of all slots that the HSM
+    /// has been initialized with.
+    ///
+    /// # Returns
+    /// * `InterfaceResult<Vec<usize>>` - A result containing a vector of slot
+    ///   identifiers available.
```
Manuthor:
```suggestion
```

crate/interfaces/src/hsm/hsm_store.rs
```diff
@@ -236,9 +237,178 @@ impl ObjectsStore for HsmStore {
...
+
+        if !check_basic_compatibility(&search_attributes, state) {
+            return Ok(uids);
+        }
```
Manuthor: It should be an error (IMO)

crate/interfaces/src/hsm/hsm_store.rs
```diff
@@ -236,9 +237,178 @@ impl ObjectsStore for HsmStore {
...
+        let Some(object_filter) = derive_hsm_object_filter(&search_attributes) else {
+            return Ok(uids);
+        };
```
Manuthor: Add `impl TryFrom<&Attributes> for HsmObjectFilter` and replace this by:

```suggestion
        let object_filter = HsmObjectFilter::try_from(&search_attributes)?;
```

crate/interfaces/src/hsm/hsm_store.rs
```diff
@@ -236,9 +237,178 @@ impl ObjectsStore for HsmStore {
...
+    }
+
+    Some(object_filter)
+}
+
+/// Extracts key size filter (if any).
+fn extract_key_size_filter(researched_attributes: &Attributes) -> Option<usize> {
+    researched_attributes
+        .cryptographic_length
+        .map(|cryptographic_length| usize::try_from(cryptographic_length).unwrap_or(0))
```
Manuthor: Maybe replace with a getter function on Attributes (crate/kmip/src/kmip_2_1/kmip_attributes.rs)

crate/interfaces/src/hsm/hsm_store.rs
```diff
@@ -236,9 +237,178 @@ impl ObjectsStore for HsmStore {
...
+/// Derives the HSM object filter from attributes.
+/// Returns `None` if incompatible combination.
+fn derive_hsm_object_filter(researched_attributes: &Attributes) -> Option<HsmObjectFilter> {
+    let mut object_filter = HsmObjectFilter::Any;
+
+    if let Some(cryptographic_algorithm) = researched_attributes.cryptographic_algorithm {
+        object_filter = match cryptographic_algorithm {
+            CryptographicAlgorithm::AES => HsmObjectFilter::AesKey,
+            CryptographicAlgorithm::RSA => HsmObjectFilter::RsaKey,
+            _ => return None,
```
Manuthor: If you don't mind, I would return an error instead of None

Manuthor: and same for other filters please

Manuthor (changes requested):
Thanks again for your contribution.
Just a few comments. Tell me if it is unclear

crate/interfaces/src/hsm/hsm_store.rs
SG-O: Based on existing `find()` implementations, it shouldn't raise an error. See:
https://github.com/Cosmian/kms/blob/b31cc1a4614a415b97fee36ae454fdb7d54cb245/crate/server_database/src/stores/sql/sqlite.rs#L753-L778
https://github.com/Cosmian/kms/blob/b31cc1a4614a415b97fee36ae454fdb7d54cb245/crate/server_database/src/stores/sql/locate_query.rs#L188-L332

crate/interfaces/src/hsm/hsm_store.rs
SG-O: If we want to match existing behavior, researched_attributes may be None. If no attributes are passed we search without attribute filters (return all HSM keys).

crate/interfaces/src/hsm/hsm_store.rs
SG-O: Same as before, to match existing behavior, a search should not fail when search attributes don't match. We just don't find anything.

crate/interfaces/src/hsm/hsm_store.rs
SG-O: I've implemented errors as requested. In the `find()` function I check for those errors and return no results should they fail to match the behavior of other 'find()' implementations (as described above).

SG-O:
@Manuthor I need your input for the open change requests.

crate/interfaces/src/hsm/hsm_store.rs
Manuthor: Thanks for digging. Indeed you are right even it remains strange.

SG-O:
I added the debug output as requested.

SG-O:
I have read the CLA Document and I hereby sign the CLA

PR #565, opened 2025-09-23
fix(enhancement): Covercrypt instantiation

### LINKED ISSUE #84: the KMS creates a new Covercrypt object at each call:
Changed but it seems that the KMS is designed to create a new Covercrypt object at each call anyway:
```rust
rekey_keypair_cover_crypt(kms, Covercrypt::default(), &owm.id, user, action, params).await
```

_Originally posted by @ackRow in https://github.com/Cosmian/kms/pull/63#discussion_r1379816736_

Manuthor:
Closes #84


Manuthor:
Closing since changes have to be done in covercrypt instead: https://github.com/Cosmian/cover_crypt/issues/165

PR #574, opened 2025-10-01
fix(bug) HSM key search fails after encountering incompatible key

### LINKED ISSUE #573: HSM key search fails to load subsequent keys after encountering incompatible key
I found a bug in the new HSM key search. When an incompatible key is encountered, it interrupts loading of subsequent keys with handles greater than the invalid key.

I already have a fix for the issue, but I am still working on adding proper tests to prevent future oversights like this.

This bug does not impact non HSM search related operations.

**Steps to Reproduce**

1. Add valid objects (AES, RSA) and also incompatible objects (like ECC keys) to the HSM.

2. Use the LOCATE command to list the objects.

SG-O:
Closes #573
Also fixes a bug that was discovered due to the reorganization of HSM tests where get_key_metadata wouldn't properly return the key label for private keys on SC HSM.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1217,6 +1217,7 @@ dependencies = [
  "async-trait",
  "cosmian_kms_interfaces",
  "cosmian_logger",
+ "futures",
  "libloading",
  "lru",
  "pkcs11-sys",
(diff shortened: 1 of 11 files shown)


Manuthor (commented):
Good job, thank you for the fixes!
Did you run the tests on your smart card Nitrokey?

SG-O:
@Manuthor I did. That's how I stumbled across that 'get_key_metadata' bug. The only platform I can't test against is Proteccio.

PR #579, opened 2025-10-07
Add option to enable automatic unwrapping for Get and Export requests

### LINKED ISSUE #578: Wrapped keys not unwrapped on Get when key_wrap_type is omitted
When performing a Get request without specifying KeyWrapType, the KMS returns a wrapped key if a key_wrapping_key is configured. 

This seems unexpected, because the KMIP spec says:
[“If a client registered a key in a given format, the server SHALL be able to return the key during the Get operation in the same format that was used when the key was registered.”](https://docs.oasis-open.org/kmip/spec/v1.4/os/kmip-spec-v1.4-os.html#_Toc490660851)

Also the current behavior breaks PyKMIP, which does not support passing a KeyWrapType in a Get request.

Is this the intended behavior and if yes, could a config option be introduced to change this?

SG-O:
Closes #578
This PR introduces a new configuration option, `default_unwrap_types`, which enables automatic unwrapping of objects for specified ObjectTypes when performing a `Get` or `Export` operation without explicitly providing a `KeyWrapType`.

Added tests for object export and unwrap behavior.

Updated documentation to describe the new configuration option.

### DIFF
--- a/crate/server/src/config/params/server_params.rs
+++ b/crate/server/src/config/params/server_params.rs
@@ -97,6 +97,14 @@ pub struct ServerParams {
     /// The Key Wrapping Key, if any
     pub key_wrapping_key: Option<String>,
 
+    /// Specifies which KMIP object types should be automatically unwrapped when retrieved
+    ///
+    /// Each entry must be the string name of a KMIP `ObjectType`, for example:
+    /// `["SecretData", "SymmetricKey"]`.
+    ///
+    /// If `None`, no automatic unwrapping will be performed.
+    pub default_unwrap_types: Option<Vec<String>>,
+
     /// The non-revocable key ID used for demo purposes
     pub non_revocable_key_id: Option<Vec<String>>,
 
@@ -203,6 +211,7 @@ impl ServerParams {
             },
             slot_passwords,
             key_wrapping_key: conf.key_encryption_key,
+            default_unwrap_types: conf.default_unwrap_type,
             non_revocable_key_id: conf.non_revocable_key_id,
             privileged_users: conf.privileged_users,
             proxy_params: ProxyParams::try_from(&conf.proxy)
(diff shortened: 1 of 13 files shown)


crate/server/src/config/params/server_params.rs
```diff
@@ -229,7 +238,9 @@ impl fmt::Debug for ServerParams {
             .field("main_db_params", &self.main_db_params)
             .field("clear_db_on_start", &self.clear_db_on_start)
             .field("unwrapped_cache_max_age", &self.unwrapped_cache_max_age)
-            .field("non_revocable_key_id", &self.non_revocable_key_id);
+            .field("non_revocable_key_id", &self.non_revocable_key_id)
+            .field("key_wrapping_key", &self.key_wrapping_key)
```
Manuthor: Remove key debug
```suggestion
```

documentation/docs/server_cli.md
```diff
@@ -162,22 +162,27 @@ Options:
           Print the server configuration information and exit
       --hsm-model <HSM_MODEL>
           The HSM model.
-          Trustway Proteccio and Utimaco General purpose HSMs are supported. [default: proteccio] [possible values: proteccio, utimaco]
+          Trustway Proteccio, Utimaco General purpose HSM, Smartcard HSM, and SoftHSM2 are supported. [default: proteccio] [possible values: proteccio, utimaco, softhsm2, smartcardhsm]--hsm-admin <HSM_ADMIN>
```
Manuthor: Something is wrong here with this autogenerated file

crate/server/src/config/params/server_params.rs
SG-O: I thought the lack of debug output for key_wrapping_key was an oversight. Will remove!

documentation/docs/server_cli.md
SG-O: I noticed this as well. That's why I removed the code block in the middle of the outer code block. That fixed the output when running locally (mkdocs serve -a 0.0.0.0:8003). 

Unrelated to this, I just noticed a duplicate `--hsm-admin <HSM_ADMIN>`. Will fix.

Manuthor (approved):

PR #601, opened 2025-11-18
Feat/azure_ekm_proxy_api_version=0.1-preview

### LINKED ISSUE #653: Support EKM Azure

HatemMn:
Issue #653 

## Overview : 
Basically an implementation of Azure EKM API as following  in a loyal manner the specifications. This is version **0.1-preview**, the code is flexible for future versions

## What's done : 

- [x] Added the necessary configs as well as the possibility to enable/disable azure EKM
- [x] Implemented the info / get metadata / Wrap / unwrap endpoints like specified
- [x] created all specified error responses, despite some of them being currently un-used
- [x] authentification ( mTLS )
- [x] Tests
- [x] About timeout : no more guards about it. The KMS does "his best"
- [x] Small refactoring with MS DKE

## Otherwise :

- [ ] the "1. Prerequisites" (read page 6) is not documented in the docs, users should refer to [azure docs](https://learn.microsoft.com/en-us/azure/key-vault/managed-hsm/) I guess

Closes #653 

## Appendix

Following are the docs diagrams
image

image

### DIFF
--- a/.github/scripts/azure_ekm_mtls_test.sh
+++ b/.github/scripts/azure_ekm_mtls_test.sh
@@ -44,8 +44,8 @@ hostname = "127.0.0.1"
 port = ${KMS_PORT}
 
 [tls]
-tls_cert_file = "${DATA_DIR}/server_cert.pem"
-tls_key_file = "${DATA_DIR}/server_key.pem"
+tls_p12_file = "${DATA_DIR}/server.p12"
+tls_p12_password = "password"
 clients_ca_cert_file = "${DATA_DIR}/client_ca_cert.pem"
 
 [db]


.cargo/config.toml
```diff
@@ -35,3 +35,7 @@ publish-dry-crate = "publish --dry-run --allow-dirty"
 #   "-C",
 #   "target_cpu=native",
 # ]
+
+# [target.x86_64-unknown-linux-gnu] # TODO: uncomment this after dev is finished
+# linker = "clang"
+# rustflags = ["-C", "link-arg=-fuse-ld=mold"]
```
Manuthor: Just a comment to remember to revert this changes

.vscode/settings.json
```diff
@@ -108,5 +108,8 @@
   "sqltools.useNodeRuntime": true,
   "rust-analyzer.cargo.features": [
     "non-fips"
-  ]
+  ],
```
Manuthor: Revert (and make it a user only configuration)

crate/crypto/src/crypto/symmetric/rfc5649.rs
```diff
@@ -116,6 +113,108 @@ pub fn rfc5649_unwrap(ciphertext: &[u8], kek: &[u8]) -> CryptoResult<Zeroizing<V
     Ok(plaintext)
 }
 
+//             // Encrypt block using AES with ECB mode i.e. raw AES as specified in
```
Manuthor: Remove

.cargo/config.toml
```diff
@@ -35,3 +35,8 @@ publish-dry-crate = "publish --dry-run --allow-dirty"
 #   "-C",
 #   "target_cpu=native",
 # ]
+
+# can increase build time for system that support mold
```
HatemMn: this was added intentionally

.vscode/settings.json
```diff
@@ -108,5 +108,8 @@
...
+  "rust-analyzer.cargo.extraEnv": {
```
HatemMn: this was also added intentionally

.cargo/config.toml
Manuthor: Could you please remove this? If needed, add an internal README with specific build instructions

.vscode/settings.json
Manuthor: Same here, remove it and add it to an internal README. IMHO, .vscode/settings.json should be removed completely.

crate/server/src/config/command_line/azure_ekm_config.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// WARNING: This bypasses mTLS authentication entirely. Only use for testing!
+    #[clap(
+        long,
+        env = "KMS_AZURE_EKM_DISABLE_CLIENT_AUTH",
+        default_value = "false"
+    )]
+    // serde does not support skipping booleans out of the box so a custom function is used
+    #[serde(skip_serializing_if = "is_false")]
+    pub azure_ekm_disable_client_auth: bool,
```
Manuthor, line 32: Is this parameter really required?
If during testing, mTLS must be disabled, I would advise to disable this TLS configuration:
```
[tls]
# Your server certificate and private key (PKCS#12 format)
tls_p12_file = "/etc/cosmian/server-cert.p12"
tls_p12_password = "your-secure-password"

# The certificate downloaded in the previous section
# This validates the client certificate presented by Azure MHSM
clients_ca_cert_file = "/etc/cosmian/mhsm-root-ca.pem"
````

Isn't it enough?

File: crate/server/src/routes/azure_ekm/error.rs [hunk 7]
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+}
+
+impl From<KmsError> for AzureEkmErrorReply {
+    fn from(e: KmsError) -> Self {
+        let status_code = e.status_code().as_u16();
+
+        // Mapping non-internal errors status numeric code to an error code string
+        let code = match status_code {
```
Manuthor, line 49: Do not convert to u16 and match directly on e.status_code()?

crate/server/src/routes/azure_ekm/handlers.rs
```diff
@@ -0,0 +1,464 @@
+#![allow(clippy::panic)]
```
Manuthor: remove this!

crate/server/src/routes/azure_ekm/handlers.rs
```diff
@@ -0,0 +1,464 @@
...
+                    let algorithm = key_block.cryptographic_algorithm().ok_or_else(|| {
+                        KmsError::ServerError("Cryptographic algorithm not set.".to_owned())
+                    })?;
+                    let key_length = key_block
+                        .cryptographic_length
+                        .ok_or_else(|| KmsError::ServerError("Key length not set.".to_owned()))?;
+                    // Check algorithm and build response
+                    match algorithm {
+                        CryptographicAlgorithm::AES => {
+                            if key_length == 256 {
```
Manuthor: Add a const variable for this 256-value and mention on it that is a Azure ProxyEKM restriction.

crate/server/src/routes/ms_dke/mod.rs
```diff
@@ -127,65 +125,42 @@ async fn internal_get_key(
...
-                        },
-                        cache: DkePublicKeyCache {
-                            expiration: {
-                                // make the key valid for one day
-                                let now = Utc::now();
-                                let later = now + Duration::days(1);
+        )
+    })?;
+    let (modulus, public_exponent) =
+        get_rsa_key_metadata_from_public_key(kms, key_id, &user).await?;
```
Manuthor, line 138: Watch out here.
Even your changes seem to be correct, there are currently no tests to validate if MS DKE encryption is broken or not.

.cargo/config.toml
HatemMn: Okay, I will look where to add it

It's not "needed" per se - but compilation takes so long (in my case? with an i9 tho) that I had to tweak it to win some seconds

crate/crypto/src/crypto/symmetric/rfc5649.rs
HatemMn: image

?

Probably an ui bug

crate/server/src/config/command_line/azure_ekm_config.rs
HatemMn: Thanks for the dicussion earlier,

For reference, we agree (on the time being) to keep this variable because (in "normal" flows) the ekm is expected to throw an error if that TLS config is disabled as you suggested

I might understand that the purpose of `azure_ekm_disable_client_auth` is not clear, I will try to further document it

It serves a similar pupose than this : 

```rust
    /// This setting turns off the validation of the tokens
    /// used by this server's Google Workspace CSE feature.
    #[clap(
        long,
        env = "KMS_GOOGLE_CSE_DISABLE_TOKENS_VALIDATION",
        default_value = "false"
    )]
    pub google_cse_disable_tokens_validation: bool,
```

crate/server/src/routes/azure_ekm/handlers.rs
HatemMn: sure, good idea

crate/server/src/routes/azure_ekm/handlers.rs
HatemMn: good catch thanks - some sort of debug artifact that slipped

crate/server/src/routes/azure_ekm/error.rs
HatemMn: image

The docs lack info on the errors "nomenclature" ( cf page 10) but they make it clear that the proxy understands those errors in that table and calls the rest "server errors" - so I managed to convert the known errors to standardised EKMReplies, and leave all the rest in the hands of the last variant

.vscode/settings.json
HatemMn: I kinda agree that we're not lacking folders to clutter the repo and it might be useful to delete this and add a "developers README" with good configs

I red the README and for now the only thing that should (probably) stay is that CSpell table (if all the team uses vscode... Otherwise we should simply agree on a shared spelling checker ( there is another file I don't like called "ltex.dictionary.en-US.txt" )

I will put an issue with all of these comments

PR #646, opened 2025-12-17
fix(RUSTSEC-2023-0071): remove rsa dependency

Manuthor:
The goal of this PR is to remove `rsa` crate dependency (and minimize required-dependencies by removing sqlx).

As reminder, the `rsa` crate dependency has the unresolved RUSTSEC-2023-0071. The reason is du to the Marvin Attack timing vulnerability in rsa crate v0.9.x. No patched/tested version available yet.

`rsa` crate is a dependency of:
- `sqlx`
- `openidconnect`
- `jwt-simple`
- `cosmian_crypto_core` v9.6

Those 4 previous dependencies are then removed in this PR.
- `openidconnect` is removed in favor of manual OIDC implementations
- `jwt-simple` is replaced by `jsonwebtoken`
- old crate`cloudproof_findex` (->crypto_core->rsa) has been removed
- `sqlx` has been replaced by those crates:
    - tokio-postgres = { version = "0.7", features = ["with-uuid-1", "with-serde_json-1"] }
    - deadpool-postgres = "0.14"
    - mysql_async = "0.36"
    - tokio-rusqlite = "0.7"
    - rusqlite = { version = "0.37", features = ["bundled", "serde_json"] }

**In addition:**
_About tests:_
- Tests under Nix environment use a pure Nix shell - to harden the test env and reuse glibc 2.27 and OpenSSL v3.1.2. For example: 
    - Utimaco and pyKmip tests have a proper Nix derivation recipe
- Enable inner cargo-tests in`nix-build`
- Fix test_kms_server when running under certificate authentication (when FIPS reuse FIPS OpenSSL version built by server)

_About database dependencies:_
- Remove the useless `SessionParams` everywhere

_About CLI:_
- Make pure FIPS compliant when needed

### DIFF
--- a/crate/cli/src/tests/kms/google_cmd/key_pairs_enabling.rs
+++ b/crate/cli/src/tests/kms/google_cmd/key_pairs_enabling.rs
@@ -1,6 +1,5 @@
 use std::process::Command;
 
-use assert_cmd::prelude::*;
 use serde::Deserialize;
 use test_kms_server::start_default_test_kms_server;
 
@@ -8,7 +7,6 @@ use crate::{
     config::COSMIAN_CLI_CONF_ENV,
     error::CosmianError,
     tests::{
-        PROG_NAME,
         kms::{google_cmd::identities::create_gmail_api_conf, utils::recover_cmd_logs},
     },
 };
@@ -53,7 +51,7 @@ fn list_keypairs(cli_conf_path: &str, user_id: &str) -> Result<ListKeyPairsRespo
         .iter()
         .map(std::string::ToString::to_string)
         .collect();
-    let mut cmd = Command::cargo_bin(PROG_NAME)?;
+    let mut cmd = Command::new("cosmian");
     cmd.env(COSMIAN_CLI_CONF_ENV, cli_conf_path);
     cmd.arg(KMS_SUBCOMMAND)
         .arg("google")
@@ -80,7 +78,7 @@ fn get_keypairs(
         .iter()
         .map(std::string::ToString::to_string)
         .collect();
-    let mut cmd = Command::cargo_bin(PROG_NAME)?;
+    let mut cmd = Command::new("cosmian");
     cmd.env(COSMIAN_CLI_CONF_ENV, cli_conf_path);
     cmd.arg(KMS_SUBCOMMAND)
         .arg("google")
@@ -107,7 +105,7 @@ fn disable_keypairs(
         .iter()
         .map(std::string::ToString::to_string)
         .collect();
-    let mut cmd = Command::cargo_bin(PROG_NAME)?;
+    let mut cmd = Command::new("cosmian");
     cmd.env(COSMIAN_CLI_CONF_ENV, cli_conf_path);
     cmd.arg(KMS_SUBCOMMAND)
         .arg("google")
@@ -132,7 +130,7 @@ fn enable_keypairs(
         .iter()
         .map(std::string::ToString::to_string)
         .collect();
-    let mut cmd = Command::cargo_bin(PROG_NAME)?;
+    let mut cmd = Command::new("cosmian");
     cmd.env(COSMIAN_CLI_CONF_ENV, cli_conf_path);
     cmd.arg(KMS_SUBCOMMAND)
         .arg("google")


crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -1,46 +1,53 @@
...
 use serde_json::Value;
-use sqlx::{
-    ConnectOptions, Executor, MySql, Pool, Row, Transaction,
-    mysql::{MySqlConnectOptions, MySqlPoolOptions, MySqlRow},
-};
 use uuid::Uuid;
 
 // Default MySQL lock wait timeout (seconds) applied to every new session.
 // MySQL default is ~50s; 10s is more appropriate for tests and reduces long stalls.
-const DEFAULT_LOCK_WAIT_TIMEOUT_SECS: u32 = 10;
```
tbrezot, line 28: Maybe remove this comment too.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -1,46 +1,53 @@
...
+    // - "ERROR 40001 (1213): Deadlock found when trying to get lock; ..."
+    // - "ER_LOCK_DEADLOCK"
+    msg.contains("Deadlock found when trying to get lock")
+        || msg.contains("(1213)")
+        || msg.contains("ER_LOCK_DEADLOCK")
+}
+
+fn mysql_deadlock_backoff_ms(attempt: u32) -> u64 {
+    // Exponential-ish backoff: 50, 100, 200, 400, 800, 1600ms
+    50_u64.saturating_mul(1_u64 << attempt.min(10))
```
tbrezot: Why not simply `50 * (1 << attempt.min(10))` ? You know that your back-off value will is bound by `50 * (1 << 10) = 51200` which does not overflow.

tbrezot: Also, didn't you mean to use `MYSQL_DEADLOCK_MAX_RETRIES` instead of 10 ?

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -98,63 +102,71 @@ impl MySqlPool {
...
-        // Create the tables if they don't exist
-        let mysql_pool = Self { pool };
-
-        // Blanket implementation of SqlMainStore for SqlDatabase
-        mysql_pool.start(clear_database).await?;
-        Ok(mysql_pool)
+        let default_conns: usize = num_cpus::get().saturating_mul(2).min(10);
+        let max_conns: usize = max_connections
+            .and_then(|v| usize::try_from(v).ok())
+            .unwrap_or(default_conns);
```
tbrezot, line 235: It seems this max connection number is always set when calling `instantiate` : thus why bothering to provide a default value ?

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -98,63 +102,71 @@ impl MySqlPool {
         clear_database: bool,
         max_connections: Option<u32>,
     ) -> DbResult<Self> {
-        let options = MySqlConnectOptions::from_str(connection_url)?
-            // disable logging of each query
-            .disable_statement_logging();
+        let opts = mysql_async::Opts::from_url(connection_url).map_err(DbError::from)?;
 
         // Default: reduce deadlocks by using READ COMMITTED isolation level per session
         // This is applied for every new connection.
```
tbrezot: Can't see anything related to READ COMMITED in this function.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -98,63 +102,71 @@ impl MySqlPool {
...
         // Also set a reasonable lock wait timeout (seconds) to fail faster under contention
```
tbrezot: Can't see anything related to setting the lock timeout in this function.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -166,52 +178,63 @@ impl SqlDatabase<MySql> for MySqlPool {
...
-        let uid = match create_(uid, owner, object, attributes, tags, &mut tx).await {
-            Ok(uid) => uid,
-            Err(e) => {
-                tx.rollback().await.context("transaction failed")?;
-                return Err(InterfaceError::Db(format!(
-                    "creation of object failed: {e}"
-                )));
+        let max_retries = MYSQL_DEADLOCK_MAX_RETRIES;
+        let mut attempt = 0_u32;
+        loop {
```
tbrezot: Consider using a `for` loop instead as it better conveys your intend and prevents the possibility of a bug causing an infinite loop.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -166,52 +178,63 @@ impl SqlDatabase<MySql> for MySqlPool {
...
+                    );
+                    if is_deadlock && attempt < max_retries {
+                        let delay_ms = mysql_deadlock_backoff_ms(attempt);
+                        attempt += 1;
+                        tokio::time::sleep(std::time::Duration::from_millis(delay_ms)).await;
+                        continue;
+                    }
+                    return Err(InterfaceError::from(e));
+                }
+            };
```
tbrezot: I would advice you to extract the core logic of the transaction such that you can handle the looping and backing-off independently:
```rust
fn create(...) -> ... {
    async fn transact(cnx: ...) -> ... {
        /* perform all operations, simply unwrapping error with ? */
    }
    for attempt in 0..MAX_ALLOWED_ATTEMPT {
        let cnx = self.get_configured_connection().await?;
        match transact(cnx).await {
            Ok(v) => return Ok(v),
            Err(e) => { /* manage error and lock */ }
        } 
    }
    Err(InterfaceError::Other("too much contention: too many attempts".to_string()))
}
```

Imho, this allows to express the transaction in a straightforward way, unencumbered of the error management which currently clouds the transaction logic.

tbrezot: As it is used several times, the backing-off logic could be factorized : you just need to take the transaction function as argument.

tbrezot (approved):
Nothing much to say, except that I think the three implementations are *not* behaviorally equivalent : 
- MySQL isolation level is set to `READ_COMMITED` while it is not the case in Postgres,
- there does not seem to be any retry logic in the postgres implementation which would lead to an error the first time high contention is encountered (arguably, Postgres lock are more fined-grained than MySQL's/SQLite's which counter-balances the difference).

This difference should be *fine*, as long as it is documented (the abstraction leaks).

crate/server_database/src/stores/sql/mysql.rs
Manuthor: indeed

crate/server_database/src/stores/sql/mysql.rs
Manuthor: I've rewritten the function following your advices.

crate/server_database/src/stores/sql/mysql.rs
Manuthor: It is always set but as an optional argument (if None, default_conns should be used)

PR #658, opened 2025-12-24
Support RFC 3394 (AESKeyWrap with **no** padding)

HatemMn:
This PR is mandatory for Azure EKM specs : https://github.com/Cosmian/kms/pull/601

Due to the nature of this update, it has been separated from the #601 in order to focus on keeping the code consistent and avoid dept

## Overview : 

- [x] Added an algorithm for the RFC3394 keywrap (/home/hatem/Desktop/repos/kms/crate/crypto/src/crypto/symmetric/rfc3394.rs )
- [x] re-coded the rfc5694 function, used openssl, added even more tests
- [x] refactored some constants
- [x] added support for 24 byte keks for the rfc 5694
- [x] Changed how the encrypt/decrypt operations "route" the AES encryption according to the block cypher mode the use choses before proceeding to encrypt with AES (check below)
- [x] fix  confusion in the codebase between the two modes

# New **BlockCipherMode** mapping : 

According to available specifications, from now on, when choosing : 
**BlockCipherMode::NISTKeyWrap -> The RFC 3394 Algorithm will be used
BlockCipherMode::AESKeyWrapPadding -> RFC 5694**

In any case, it is preffered to avoid using  3394 as it's deprecated in favor of the 5694

# ⚠️ WARNING (must be included in release notes!) ⚠️ 

Previously saved keys in a database will still be compatible and nothing has to be done, the code will fix their blockCipherMode to the new mapping if set

**However** , any previously **manually** exported key in the JSON format should be manually fixed. This can be trivially done using the following command : `sed -i 's/NISTKeyWrap/AESKeyWrapPadding/g' your_exported_key.json`

### DIFF
--- a/crate/crypto/src/crypto/symmetric/rfc3394.rs
+++ b/crate/crypto/src/crypto/symmetric/rfc3394.rs
@@ -50,7 +50,7 @@ pub fn rfc3394_wrap(plaintext: &[u8], kek: &[u8]) -> CryptoResult<Vec<u8>> {
     // Allocate output buffer: wrapped size is plaintext + 8 bytes (IV) + 2 extra blocks for cipher_final.
     // The extra blocks will not propagate to the result as it's truncated to the actual size. Due to how the openssl library is programmed,
     // not adding at least 1 extra block results in a panic. We chose to add two because that's how the openssl library operates when using this cipher.
-    let mut ciphertext = vec![0_u8; n_bytes + AES_WRAP_BLOCK_SIZE + AES_BLOCK_SIZE];
+    let mut ciphertext = vec![0_u8; n_bytes + AES_WRAP_BLOCK_SIZE + (AES_BLOCK_SIZE * 2)];
 
     // Perform the key wrap operation
     let mut written = ctx.cipher_update(plaintext, Some(&mut ciphertext))?;
@@ -81,7 +81,10 @@ pub fn rfc3394_unwrap(ciphertext: &[u8], kek: &[u8]) -> CryptoResult<Zeroizing<V
     ctx.decrypt_init(Some(cipher), Some(kek), None)?;
 
     // Allocate output buffer: unwrapped size is ciphertext - 8 bytes (IV) + extra blocks for cipher_final. Same comments as above.
-    let mut plaintext = Zeroizing::new(vec![0_u8; n_bytes - AES_WRAP_BLOCK_SIZE + AES_BLOCK_SIZE]);
+    let mut plaintext = Zeroizing::new(vec![
+        0_u8;
+        n_bytes - AES_WRAP_BLOCK_SIZE + (AES_BLOCK_SIZE * 2)
+    ]);
 
     // Perform the key unwrap operation
     let mut written = ctx.cipher_update(ciphertext, Some(&mut plaintext))?;
--- a/crate/crypto/src/crypto/symmetric/rfc5649.rs
+++ b/crate/crypto/src/crypto/symmetric/rfc5649.rs
@@ -66,7 +66,7 @@ pub fn rfc5649_wrap(plaintext: &[u8], kek: &[u8]) -> CryptoResult<Vec<u8>> {
     };
 
     // Allocate output buffer with extra space for cipher_final
-    let mut ciphertext = vec![0_u8; padded_len + AES_BLOCK_SIZE];
+    let mut ciphertext = vec![0_u8; padded_len + (AES_BLOCK_SIZE * 2)];
 
     // Perform the key wrap operation
     let mut written = ctx.cipher_update(plaintext, Some(&mut ciphertext))?;
@@ -97,7 +97,10 @@ pub fn rfc5649_unwrap(ciphertext: &[u8], kek: &[u8]) -> CryptoResult<Zeroizing<V
 
     // Allocate output buffer: maximum plaintext size is ciphertext - 8 bytes (AIV)
     // Add extra space for cipher_final
-    let mut plaintext = Zeroizing::new(vec![0_u8; n_bytes - AES_WRAP_BLOCK_SIZE + AES_BLOCK_SIZE]);
+    let mut plaintext = Zeroizing::new(vec![
+        0_u8;
+        n_bytes - AES_WRAP_BLOCK_SIZE + (AES_BLOCK_SIZE * 2)
+    ]);
 
     // Perform the key unwrap operation
     let mut written = ctx.cipher_update(ciphertext, Some(&mut plaintext))?;
--- a/crate/kmip/src/kmip_0/kmip_types.rs
+++ b/crate/kmip/src/kmip_0/kmip_types.rs
@@ -765,49 +765,34 @@ pub enum ShreddingAlgorithm {
     Unsupervised = 0x3,
 }
 
-#[allow(deprecated)] // this is mandatory since we do want to signal that variant is deprecated to the "user" but still keep it
-// this module was created because it's not possible to `allow` lints and use `kmip_enum` at the same time. The module is immediately
-// re-exported below, it only exists for proper compilation and has 0 effects on the final executable
-mod block_cipher_mode {
-    use super::{
-        Deserialize, Display, KmipEnumDeserialize, KmipEnumSerialize, Serialize, kmip_enum,
-    };
-
-    #[kmip_enum]
-    pub enum BlockCipherMode {
-        CBC = 0x0000_0001,
-        ECB = 0x0000_0002,
-        PCBC = 0x0000_0003,
-        CFB = 0x0000_0004,
-        OFB = 0x0000_0005,
-        CTR = 0x0000_0006,
-        CMAC = 0x0000_0007,
-        CCM = 0x0000_0008,
-        GCM = 0x0000_0009,
-        CBCMAC = 0x0000_000A,
-        XTS = 0x0000_000B,
-        AESKeyWrapPadding = 0x0000_000C, // RFC 5649
-        NISTKeyWrap = 0x0000_000D,       // RFC 3394
-        X9102AESKW = 0x0000_000E,
-        X9102TDKW = 0x0000_000F,
-        X9102AKW1 = 0x0000_0010,
-        X9102AKW2 = 0x0000_0011,
-        AEAD = 0x0000_0012,
-        // Extensions - 8XXXXXXX
-        // AES GCM SIV
-        GCMSIV = 0x8000_0002,
-        // This variant was introduced to support backward compatibility with versions prior to 5.15
-        // In the database layer, right after deserialization, objects that have a saved BlockCipherMode (via their `KeyWrappingData`) are tested for this mode and
-        // converted to AESKeyWrapPadding if found.
-        #[deprecated(
-            since = "5.15.0",
-            note = "Use `AESKeyWrapPadding` instead. This value exists only for backward compatibility and will be removed in a future version."
-        )]
-        LegacyNISTKeyWrap = 0x8000_000D,
-    }
+#[kmip_enum]
+pub enum BlockCipherMode {
+    CBC = 0x0000_0001,
+    ECB = 0x0000_0002,
+    PCBC = 0x0000_0003,
+    CFB = 0x0000_0004,
+    OFB = 0x0000_0005,
+    CTR = 0x0000_0006,
+    CMAC = 0x0000_0007,
+    CCM = 0x0000_0008,
+    GCM = 0x0000_0009,
+    CBCMAC = 0x0000_000A,
+    XTS = 0x0000_000B,
+    AESKeyWrapPadding = 0x0000_000C, // RFC 5649
+    NISTKeyWrap = 0x0000_000D,       // RFC 3394
+    X9102AESKW = 0x0000_000E,
+    X9102TDKW = 0x0000_000F,
+    X9102AKW1 = 0x0000_0010,
+    X9102AKW2 = 0x0000_0011,
+    AEAD = 0x0000_0012,
+    // Extensions - 8XXXXXXX
+    // AES GCM SIV
+    GCMSIV = 0x8000_0002,
+    // This variant was introduced to support backward compatibility with versions prior to 5.15
+    // In the database layer, right after deserialization, objects that have a saved BlockCipherMode (via their `KeyWrappingData`) are tested for this mode and
+    // converted to AESKeyWrapPadding if found.
+    LegacyNISTKeyWrap = 0x8000_000D,
 }
-
-pub use block_cipher_mode::BlockCipherMode;
 
 /// Padding Method Enumeration
 #[kmip_enum]
--- a/crate/server_database/src/lib.rs
+++ b/crate/server_database/src/lib.rs
@@ -65,7 +65,6 @@ use cosmian_kmip::kmip_2_1::kmip_objects::Object;
 ///
 /// This function checks if the object has a `KeyWrappingData` with the legacy
 /// `NISTKeyWrap` value and converts it to the correct `AESKeyWrapPadding` value.
-#[allow(deprecated)] // Allow use of LegacyNISTKeyWrap for backward compatibility migration
 pub(crate) fn migrate_block_cipher_mode_if_needed(mut object: Object) -> Object {
     use cosmian_kmip::{
         kmip_0::kmip_types::BlockCipherMode,


crate/kmip/src/kmip_0/kmip_types.rs
```diff
@@ -780,7 +780,7 @@ pub enum BlockCipherMode {
     CBCMAC = 0x0000_000A,
     XTS = 0x0000_000B,
     AESKeyWrapPadding = 0x0000_000C,
-    NISTKeyWrap = 0x8000_000D,
+    NISTKeyWrap = 0x8000_000D, // FIXME
```
HatemMn: This comment is left here intentionally

crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
```diff
@@ -185,7 +199,12 @@ impl SymCipher {
             Self::Aes256Gcm => Ok(Cipher::aes_256_gcm()),
             Self::Aes128Xts => Ok(Cipher::aes_128_xts()),
             Self::Aes256Xts => Ok(Cipher::aes_256_xts()),
-            Self::Rfc5649_16 | Self::Rfc5649_32 => {
+            Self::Rfc3394_16 | Self::Rfc3394_24 | Self::Rfc3394_32 => {
+                crypto_bail!(CryptoError::NotSupported(
```
Manuthor, line 203: display the variant value in error message

crate/crypto/src/crypto/symmetric/rfc3394.rs
```diff
@@ -0,0 +1,210 @@
...
+        return Err(CryptoError::InvalidSize(
+            "The ciphertext size should be >= 24 and a multiple of 8".to_owned(),
+        ));
+    }
+
+    let cipher = select_cipher(kek)?;
+
+    // Initialize cipher context for decryption
+    let mut ctx = CipherCtx::new()?;
+    ctx.set_flags(CipherCtxFlags::FLAG_WRAP_ALLOW); // For some reason the code works without this flag, but let's set it anyway
```
Manuthor: if not needed, remove this line

crate/cli/src/actions/kms/symmetric/decrypt.rs
```diff
@@ -384,13 +384,11 @@ impl DecryptAction {
         });
 
         // recover the DEK
-        unwrap_key_block(dek_object.key_block_mut()?, &unwrapping_key)?;
+        unwrap_key_block(dek_object.key_block_mut()?, &unwrapping_key)?; // TODO error handling
```
Manuthor: Handle error and revert comment

.vscode/settings.json
```diff
@@ -108,5 +108,6 @@
   "sqltools.useNodeRuntime": true,
   "rust-analyzer.cargo.features": [
     "non-fips"
-  ]
+  ],
+  "rust-analyzer.cargo.targetDir": "./target/rust-analyzer",
```
Manuthor: revert

crate/crypto/src/crypto/symmetric/rfc3394.rs
```diff
@@ -0,0 +1,210 @@
...
+        return Err(CryptoError::InvalidSize(
+            "The plaintext size should be >= 16 and a multiple of 8".to_owned(),
+        ));
+    }
+
+    let cipher = select_cipher(kek)?;
+
+    // Initialize cipher context for encryption
+    let mut ctx = CipherCtx::new()?;
+    ctx.set_flags(CipherCtxFlags::FLAG_WRAP_ALLOW); // For some reason the code works without this, but it should be set anyway
```
Manuthor: Remove the line if not needed

crate/crypto/src/crypto/symmetric/rfc5649.rs
```diff
@@ -1,294 +1,121 @@
...
 
-    // Number of 64-bit blocks (block size for RFC 5649).
-    let n = n / AES_WRAP_PAD_BLOCK_SIZE;
+    let cipher = select_cipher_pad(kek)?;
 
-    // ICR stands for Integrity Check Register initially containing the IV.
-    let mut icr = iv.unwrap_or(DEFAULT_IV);
-    let mut blocks = Vec::with_capacity(n);
+    let mut ctx = CipherCtx::new()?;
+    ctx.set_flags(CipherCtxFlags::FLAG_WRAP_ALLOW); // For some reason the code works without this, but it should should anyway
```
Manuthor:
```suggestion
```

crate/crypto/src/crypto/symmetric/rfc5649.rs
```diff
@@ -297,30 +124,15 @@ mod tests {
...
-        );
-        assert_eq!(
-            rfc5649_unwrap(&wrapped_key, kek).expect("Fail to unwrap"),
-            Zeroizing::from(key_to_wrap.to_vec())
-        );
+        // Empty plaintext should fail
+        let empty: &[u8] = &[];
+        rfc5649_wrap(empty, kek).unwrap_err();
 
+        // The rest of the round-trips shouldsucceedd, including the single byte one
```
Manuthor:
```suggestion
        // The rest of the round-trips should succeed, including the single byte one
```

crate/crypto/src/crypto/wrap/unwrap_key.rs
```diff
@@ -294,6 +295,7 @@ fn unwrap_with_symmetric_key(
         aes_gcm_unwrap(ciphertext, &unwrap_secret)
     } else {
         // unwrap using rfc_5649
+        trace!("asbandoooooooooooooooooooooooo");
```
Manuthor: revert

crate/kmip/src/kmip_0/kmip_types.rs
Manuthor: You can fix it to `0x0000_000D`

crate/crypto/src/crypto/symmetric/rfc3394.rs
```diff
@@ -0,0 +1,210 @@
...
+    ctx.encrypt_init(Some(cipher), Some(kek), None)?;
+
+    // Allocate output buffer: wrapped size is plaintext + 8 bytes (IV) + 2 extra blocks for cipher_final
+    // The extra blocs will not propagate to the result as its truncated to the actual size. Due to how the openssl library is programmed,
+    // not adding at least 1 extra bloc results in a panic. We chose to add two because that's how openssl library operates when using this cypher
+    let mut ciphertext = vec![0_u8; n_bytes + AES_WRAP_BLOCK_SIZE + (AES_BLOCK_SIZE * 2)];
```
tbrezot, line 53: It seems the RFC specifies that the output of the wrapping is `n+1` blocks of 64 bits. In practice, I guess the OpenSSL implementation must write full aes blocks which could explain that allocating `n_bytes + AES_WRAP_BLOCK_SIZE` causes the Rust wrapper to panic. I have to say the OpenSSL documentation is rather lacking, but I can avoid the panic by allocating `n_bytes + AES_BLOCK_SIZE` instead of the current `n_bytes + AES_WRAP_BLOCK_SIZE + 2 * AES_BLOCK_SIZE` (i.e. `AES_WRAP_BLOCK_SIZE + AES_BLOCK_SIZE` less allocated bytes). Same goes for the `unwrap` procedure.

The current version is not incorrect, but the allocated sizes raise questions and unnecessarily stresses the allocator.

crate/crypto/src/crypto/symmetric/rfc5649.rs
```diff
@@ -1,294 +1,121 @@
...
+        24 => Cipher::aes_192_wrap_pad(),
+        32 => Cipher::aes_256_wrap_pad(),
+        _ => {
             return Err(CryptoError::InvalidSize(
-                "The ciphertext is invalid. Decrypted IV is not appropriate".to_owned(),
+                "The KEK size should be 16, 24 or 32 bytes".to_owned(),
             ));
         }
+    })
+}
```
tbrezot, line 32: This could be shared with RFC3394.

crate/crypto/src/crypto/symmetric/rfc5649.rs
```diff
@@ -1,294 +1,121 @@
...
-/// Wrap a plain text of a 64-bits modulo size according to RFC 3394.
-///
-/// The function name matches the one used in the RFC and has no link to the
-/// unwrap function in Rust.
-fn wrap_64(plain: &[u8], kek: &[u8], iv: Option<u64>) -> Result<Vec<u8>, CryptoError> {
-    let n = plain.len();
 
-    if !n.is_multiple_of(AES_WRAP_PAD_BLOCK_SIZE) {
+    // Check maximum length (2^32 - 1 bytes as per NIST SP 800-38F Section 5.3.1)
+    if n_bytes >= (1_usize << 32) {
```
tbrezot: Use a `u64` as `1_usize << 32` may be equal to `0` on 32-bit architecture.

crate/kmip/src/kmip_0/kmip_types.rs
HatemMn: https://github.com/Cosmian/kms/issues/662

I don't think it's this easy, this is what I was working on this morning actually. In the database, some wrapped keys **might** save the blockCipherMode, it's pretty low risk to cause some problem if we just change it  and move on but it introduces the possibility that someone, somewhere **might** unwrap his keys with a bugged key which will cause an almost unsolvable deserialization bug

I suggest to work on this on a separate pr, unless you think otherwise

crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
HatemMn:
```rust
            Self::Rfc3394_16 | Self::Rfc3394_24 | Self::Rfc3394_32 => {
                crypto_bail!(CryptoError::NotSupported(
                    "RFC3394 is not supported in this version of openssl".to_owned()
                ))
            }
            Self::Rfc5649_16 | Self::Rfc5649_24 | Self::Rfc5649_32 => {
                crypto_bail!(CryptoError::NotSupported(
                    "RFC5649 is not supported in this version of openssl".to_owned()
                ))
            }
```

Isn't it done ?

crate/crypto/src/crypto/symmetric/rfc3394.rs
HatemMn: done

crate/cli/src/actions/kms/symmetric/decrypt.rs
HatemMn: the error was handled, pointless comment that I have already deleted

crate/crypto/src/crypto/symmetric/rfc3394.rs
HatemMn: I totally agree with what you said and have reached the same conclusions when I have coded this function, however, due to the lack of computation impact of allocating the extra block, I have chosen to write `(AES_BLOCK_SIZE * 2)` because on the openSSL source code, more specefically on the unit/integration tests, this is what they do and that code is our only reference on the matter...

Since the code is now tested in a very extensive manner, I will check what happens if I take out the `* 2` and comment back on this

crate/crypto/src/crypto/symmetric/rfc5649.rs
HatemMn: Actually they're not the same :

```rust
        16 => Cipher::aes_128_wrap(),
```

and 



```rust
        16 => Cipher::aes_128_wrap_pad(), // <----- "pad"
```

crate/crypto/src/crypto/symmetric/rfc5649.rs
HatemMn: done

crate/crypto/src/crypto/symmetric/rfc3394.rs
HatemMn: no bugs on sight, I think the only case that needs that block is the minimal case, which is tested, so i can delete it

crate/kmip/src/kmip_0/kmip_types.rs
Manuthor: I think this is wrong or partially wrong: 
- the enum value NISTKeyWrap is incorrect and we must conform to KMIP specifications in any case (we are not the only ones to follow the specs and we need to maximize integration with other products)
- we are not going to support any migration procedure from a wrong-state (incorrect value enum) to a correct state (especially not in Rust source code in any case)
- for existing wrapping keys, we are going to document this change and give the required commands to modify wrapping-key with an old blockCipherMode

crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
Manuthor: What about the key size?

HatemMn: Ah all right ! Sorry I didn't understand the original message

In the original code, no variant was provided on the error message, do you really think this is necessary ? The whole algorithm isn't supported anyway so key sizes don't matter IMO

By the way, please notice this issue : https://github.com/Cosmian/kms/issues/660

Feel free to comment it if you have an idea about that (I don't think we'll work on it now, but I think it's a bit odd to crypto_bail as "unsupported" something that we actually support

crate/kmip/src/kmip_0/kmip_types.rs
```diff
@@ -765,32 +765,50 @@ pub enum ShreddingAlgorithm {
...
-    X9102AESKW = 0x0000_000E,
-    X9102TDKW = 0x0000_000F,
-    X9102AKW1 = 0x0000_0010,
-    X9102AKW2 = 0x0000_0011,
-    AEAD = 0x0000_0012,
-    // Extensions - 8XXXXXXX
-    // AES GCM SIV
-    GCMSIV = 0x8000_0002,
+#[allow(deprecated)] // this is mandatory since we do want to signal that variant is deprecated to the "user" but still keep it
+// this module was created because it's not possible to `allow` lints and use `kmip_enum` at the same time. The module is immediatly
```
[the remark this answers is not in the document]
HatemMn: _GithubGrammarPilot_

HatemMn (commented):
I think work here is probably over, I will rebase azure EKM on this

The last modifications are squashed on this commit to ease review https://github.com/Cosmian/kms/pull/658/commits/01acb191f91ad518ec468640783d14e137803275

Possible to read the commit message to see what's new (reviews fixes + ui fixes + migrate fixes + a test)

crate/crypto/src/crypto/symmetric/symmetric_ciphers.rs
Manuthor: I believe we could handle issue #660 now (when you are ready). If you agree, create a new PR starting from this one (#658) and we'll wait that issue #660 is completely done before merging #658 to develop.

crate/kmip/src/kmip_0/kmip_types.rs
```diff
@@ -765,32 +765,50 @@ pub enum ShreddingAlgorithm {
...
+        // Extensions - 8XXXXXXX
+        // AES GCM SIV
+        GCMSIV = 0x8000_0002,
+        // This variant was introduced to support backward compatibility with versions prior to 5.15
+        // In the database layer, right after deserialization, objects that have a saved BlockCipherMode (via their `KeyWrappingData`) are tested for this mode and
+        // converted to AESKeyWrapPadding if found.
+        #[deprecated(
+            since = "5.15.0",
+            note = "Use `AESKeyWrapPadding` instead. This value exists only for backward compatibility and will be removed in a future version."
+        )]
```
Manuthor: Please remove the deprecated annotation (we support 3DES, DSA, etc. and we will not maintain deprecated annotations everywhere it has to be).

crate/server_database/src/lib.rs
```diff
@@ -55,3 +55,57 @@ pub mod reexport {
...
+use cosmian_kmip::kmip_2_1::kmip_objects::Object;
+
+/// Upgrades wrongly serialized `BlockCipherMode::LegacyNISTKeyWrap` (`0x8000_000D`) to
+/// `BlockCipherMode::AESKeyWrapPadding` (`0x0000_000C`) for backward compatibility
+/// with versions prior to 5.15.x Calling this right after deserializing ensures that no faulty
+/// data even gets read from a (previously) valid database.
+///
+/// This function checks if the object has a `KeyWrappingData` with the legacy
+/// `NISTKeyWrap` value and converts it to the correct `AESKeyWrapPadding` value.
+#[allow(deprecated)] // Allow use of LegacyNISTKeyWrap for backward compatibility migration
```
Manuthor: A comment to not forget to remove

crate/server_database/src/stores/redis/objects_db.rs
```diff
@@ -152,7 +155,7 @@ impl ObjectsDB {
             })?;
             Nonce::new(&mut *rng)
         };
-        let ct = self.dem.encrypt(
+        let ct: Vec<u8> = self.dem.encrypt(
```
Manuthor, line 158: really needed?

HatemMn: no thx for the notice it seems correctly infered

PR #681, opened 2026-01-09
feat: AWS KMS Byok

### LINKED ISSUE #650: Support of AWS BYOK

### LINKED ISSUE #685: Azure Byok might not re-wrap already wrapped in-base keys
On this file : https://github.com/Cosmian/kms/blob/9c11fc416581eb96ce8433dae5260b54ea394450/crate/cli/src/actions/kms/azure/byok/export_byok.rs

Line 73 : 
```rust
        // Export the key wrapped with the KEK
        // export the object
        let export_params = ExportObjectParams {
            unwrap: true,
            wrapping_key_id: Some(&self.kek_id),
            allow_revoked: false,
            // rest of code
        };
```

However, it states here : https://github.com/Cosmian/kms/blob/3445d47eca72d39d23491d24841c7dea770489bd/crate/kms_client/src/export_utils.rs that : 

```rust
#[derive(Default)]
pub struct ExportObjectParams<'a> {
    ///  Unwrap the object if it is wrapped
    pub unwrap: bool,
    ///  The wrapping key id to wrap the key, may be the PKCS#12 password. `wrapping_key_id` is ignored if `unwrap` is true
    pub wrapping_key_id: Option<&'a str>,
    /// `allow_revoked` - Allow the export of a revoked object
    pub allow_revoked: bool,
    /// `key_format_type` - The key format for export
    pub key_format_type: Option<KeyFormatType>,
    /// `encode_to_ttlv` - if wrapping, Encode the Key Material to JSON TTLV before wrapping
    pub encode_to_ttlv: bool,
    /// `cryptographic_parameters` - The cryptographic parameters for wrapping
    pub wrapping_cryptographic_parameters: Option<CryptographicParameters>,
    /// `authenticated_encryption_additional_data` - Wrapping using GCM mode, additional data used for encryption
    pub authenticated_encryption_additional_data: Option<String>,
}
```

So basically if someone tries to "byok" some key that has already been wrapped inside the KMS.

If he does, our KMS will (to my understanding) export the key as-is, without using the cloud provider (in this case Azure)'s KEK

And this will make decyphering fail on the other KMS

## Solutions : 
IMO the easiest would be to set `unwrap` to false, this will eliminate any possible error

Another solution would be unwrapping any previously wrapped keys before this export. This isn't hard but with can be done manually by users and I see no motive to actually do it during the BYOK operation unless asked-for

To go faster, this can be linked to https://github.com/Cosmian/kms/pull/681

HatemMn:
- [x] Introduce 2 cli actions to permit BYOK flows with AWS KMS (docs below)
- [x] Create full flow integration tests - to avoid static mocks, utility functions that mimic (a correct) AWS KMS were created to run locally using openssl

- [x] Fix **stdout** console formatting when an export action happens

## Checklist 

- [x] should succeed some manual E2E tests from cli
- [x] and some tests from the ui ( E2E )
- [x] delete the chinese algorithm and its instances
- [x] test **valid** combinations with the aws cli
- [x] proper docs on docs.cosmian

## Bonus : 
added some scripts that automate the aws byok flow, they are placed with the docs and avalable to download

#### References :

Creating keys on amazon kms : https://docs.aws.amazon.com/kms/latest/developerguide/importing-keys-conceptual.html

Importing keys https://docs.aws.amazon.com/kms/latest/developerguide/importing-keys.html

Requirements for key material :https://docs.aws.amazon.com/kms/latest/developerguide/importing-keys-conceptual.html#importing-keys-material-requirements

Closes #685 
Closes #650


HatemMn:
@copilot search this PR for "TODO" comments, useless comments, or stuff I have "forgot" around

crate/cli/src/actions/kms/aws/byok/import_kek.rs
```diff
@@ -0,0 +1,111 @@
...
+            decoded.len()
+        ));
+    }
+    Ok(s.to_owned())
+}
+
+/// Import an AWS Key Encryption Key (KEK) into the KMS.
+#[derive(Parser)]
+#[clap(verbatim_doc_comment)]
+#[clap(group(ArgGroup::new("kek_input").required(true).args(["kek_base64", "kek_file"])))] // At least one of kek_file or kek_blob must be provided
```
Manuthor:
```suggestion
#[clap(group(ArgGroup::new("kek_input").required(true).args(["kek_base64", "kek_file"])))] // At least one of kek_file or kek_base64 must be provided
```

crate/cli/src/actions/kms/aws/byok/import_kek.rs
```diff
@@ -0,0 +1,111 @@
...
+        group = "kek_input"
+    )]
+    pub(crate) kek_base64: Option<String>,
+
+    /// In case of KEK provided as a file blob.
+    #[clap(short = 'f', long, group = "kek_input")]
+    pub(crate) kek_file: Option<PathBuf>,
+
+    #[clap(short = 'w', long, required = true)]
+    pub(crate) wrapping_algorithm: AwsKmsWrappingAlgorithm,
```
Manuthor, line 57: Please move it to ExportAwsByok and add a comment for this arg with expected algorithmes (comment starting with `///`)

crate/cli/src/actions/kms/azure/byok/export_byok.rs
```diff
@@ -71,7 +71,7 @@ impl ExportByokAction {
         // Export the key wrapped with the KEK
         // export the object
         let export_params = ExportObjectParams {
-            unwrap: true,
+            unwrap: false, // `wrapping_key_id` is ignored if `unwrap` is true
```
Manuthor, line 74: Mmmh good catch.

crate/cli/src/actions/kms/aws/byok/export_key_material.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        if !tags.contains(&"aws".to_owned()) {
+            return Err(KmsCliError::InconsistentOperation(kek_tag_error(
+                "missing `aws` tag",
+            )));
+        }
+
+        let key_arn = tags.iter().find_map(|t| t.strip_prefix("key_arn:"));
+
+        let wrapping_algorithm_str = tags
```
Manuthor, line 76: I would rather move this argument to ExportByokAction:
```
    #[clap(short = 'w', long, required = true)]
    pub(crate) wrapping_algorithm: AwsKmsWrappingAlgorithm,
```
Avoiding this additional wrapping_algorithm-tag and this string conversion.

crate/cli/src/actions/kms/aws/byok/import_kek.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// The Amazon Resource Name (key ARN) of the KMS key. It's recommended to provide it for an easier export later.
+    #[clap(short = 'a', long, required = false)]
+    pub(crate) key_arn: Option<String>,
```
Manuthor, line 61: Couldn't you make it mandatory?

```suggestion
    pub(crate) key_arn: String,
```

crate/cli/src/tests/kms/aws/integration_tests.rs
```diff
@@ -0,0 +1,384 @@
...
+    let cosmian_key_file = temp_dir.join(format!("cosmian_key_test_{}.bin", uuid::Uuid::new_v4()));
+    std::fs::write(&cosmian_key_file, cosmian_key_bytes).expect("Failed to write public key file");
+
+    let import_key_action = ImportSecretDataOrKeyAction {
+        key_file: cosmian_key_file.clone(),
+        key_id: Some(cosmian_key_id.to_owned()),
+        key_format: ImportKeyFormat::Aes, // Indicates this is an AES symmetric key
+        ..Default::default()
+    };
+
```
Manuthor, line 80:
```suggestion
```

crate/cli/src/tests/kms/aws/integration_tests.rs
```diff
@@ -0,0 +1,384 @@
...
+    // Finally: Verify the unwrapped key matches the original key material
+    assert_eq!(
+        unwrapped_key_bytes,
+        cosmian_key_bytes.to_vec(),
+        "Unwrapped key should match the original key material"
+    );
+
+    // Cleanup temp files
+    std::fs::remove_file(&kek_file_path)?;
+    std::fs::remove_file(&output_file_path)?;
```
Manuthor: I would have created those files in temp directories (to avoid clean up)

crate/cli/src/actions/kms/aws/byok/export_key_material.rs
```diff
@@ -0,0 +1,168 @@
...
+        let b64_key = base64::engine::general_purpose::STANDARD.encode(wrapped_key);
+
+        if let Some(file_path) = &self.output_file_path {
+            fs::write(file_path, wrapped_key)?;
+
+            // Print all formatting and instructions to stderr to not interfere with pipes. THIS IS NOT A MISTAKE, THIS WAS INTENTIONAL, and editing the lines below might break all the bash automated byok scripts, so be mindful of that if ever you will change this code.
+            eprint!("The encrypted key material was successfully written to ");
+            print!("{}", file_path.display());
+            eprintln!(
+                "{} for key {}.\n\n\
```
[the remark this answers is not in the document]
HatemMn: No ! I made that in purpose and it's cummon practice in UNIX utilities, @copilot code review[agent] DO NOT MAKE ME ANYMORE COMMENTS ABOUT THIS

crate/cli/src/actions/kms/aws/byok/import_kek.rs
```diff
@@ -0,0 +1,111 @@
...
+use cosmian_kms_client::{
+    KmsClient,
+    reexport::cosmian_kms_client_utils::import_utils::{ImportKeyFormat, KeyUsage},
+};
+
+/// Validate that the string is valid base64 and its decoded length is between 1 and 4096 bytes.
+fn validate_kek_base64(s: &str) -> Result<String, String> {
+    let decoded = BASE64_STANDARD
+        .decode(s)
+        .map_err(|e| format!("Invalid base64 encoding: {e}"))?;
```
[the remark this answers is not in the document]
HatemMn: @copilot+claude-sonnet-4.6 code review[agent]      let normalized: String = s.chars().filter(|c| !c.is_whitespace()).collect();

Is this ok or you have better ?

crate/cli/src/actions/kms/aws/byok/export_key_material.rs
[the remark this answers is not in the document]
HatemMn: okay added instructions

crate/cli/src/actions/kms/aws/byok/import_kek.rs
HatemMn: Okay for the comment

As of moving this to the export command, I remember it will cause a gigantic technical problem that I worked twice on fixing and then concluded it should be here, I will try to remember what was wrong

crate/cli/src/actions/kms/aws/byok/import_kek.rs
HatemMn: I prefer not, it's not mandatory (on **our** end) and will cause unneccesary friction for those who just want to get a key in and a key out (like the CLI automatic scripts especially)

crate/cli/src/tests/kms/aws/integration_tests.rs
HatemMn: ?


Might be a typo

crate/cli/src/actions/kms/aws/byok/import_kek.rs
HatemMn: Okay so after revision of the code, AWS insures that a kek is **tied** to a specific algorithm. It does not make any sense to use it with another algorithm and there is no one-to-many relashionship like other kek wrap/unwrap classic flows.

If I were to put it on the export, I would be still obliged to have the tag there in order to prevent (murphy's law) people from doing unwanted combinations, and that requires a tag I will later parse during the export

I found myself parsing the tag no matter where the `pub(crate) wrapping_algorithm: AwsKmsWrappingAlgorithm,` so I considered that it's more convenient to put it on the import. Even from an user perspective it's more intuitive to specify the parameters from the beginning, and from our end it prohibits unwanted usages with the only overhead being a tag string parsing `export_key_material.rs` line 76


Please tell me what you think. We can move it, I guess, but it feels cleaner this way (and makes the export pretty fast, half parameters are optionnal)

Manuthor: Thank you for the explanation. Then your design is better and this comment is resolved.

HatemMn:
Opened this issue : https://github.com/Cosmian/kms/issues/757

Since those tests are not urgent, might be considered to merge this is everything is ok

HatemMn (commented):
test

crate/cli/src/tests/kms/aws/integration_tests.rs
HatemMn: I didn't know about this syntax before : 
```
    let tmp_dir = tempfile::TempDir::new()?;
    let temp_dir = tmp_dir.path();
 ```
 
 I used it it in my code - thanks !

File: crate/cli/src/actions/kms/aws/byok/export_key_material.rs [hunk 10]
```diff
@@ -0,0 +1,168 @@
...
+         To import into AWS KMS using the API, run:\n\
+         aws kms import-key-material \\\n\
+             --key-id {} \\\n\
+             --encrypted-key-material fileb://{} \\\n\
+             --import-token fileb://{} \\\n\
+             --expiration-model KEY_MATERIAL_DOES_NOT_EXPIRE",
+                wrapped_key.len(),
```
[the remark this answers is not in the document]
HatemMn: FOR THE LAST TIME DO NOT COMMENT THIS ANYMORE PLEASE

crate/cli/src/actions/kms/aws/byok/import_kek.rs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        let import_action = ImportSecretDataOrKeyAction {
+            key_file: match (&self.kek_file, &self.kek_base64) {
+                (Some(file), _) => file.clone(),
+                (None, Some(base64_str)) => {
+                    let temp_path = std::env::temp_dir().join(format!("{}", uuid::Uuid::new_v4()));
+                    std::fs::write(&temp_path, BASE64_STANDARD.decode(base64_str)?)?;
+                    temp_path
+                }
```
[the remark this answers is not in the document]
HatemMn: The idea is cool but we don't want to add new dependencies, drop it

PR #690, opened 2026-01-20
feat: provide /health endpoint

### LINKED ISSUE #672: Provide /health endpoint
Currently there is `/version`  GET endpoint from which we can get version information.
We also want additional information about Cosmian service - statuses of crucial components that needs to run in order for Cosmian service to run correctly.

### Environment
- Cosmian KMS version: <= 5.14
- any relation database + redis
- maybe cache information - if it is working correctly, how many elements are present in the cache,...
- Using Cosmian KMS **GET** endpoints: `{main url}/health` for health
- Using Cosmian KMS **GET** endpoints version (already present): `{main url}/version`

### Description

We want to have `/health` endpoint to tell us the running statuses of Cosmain KMS service underlying depended libraries/services.
This is important if we have load-balancer in front of vertical scaling Cosmian services. Load balancer can check `/health` endpoint for route decisions (check multiple services if they ar **UP** or **DOWN**).  Also important, if other services relays/depends on Cosmian KMS service - other service can include Cosmian KMS `/health` endpoint for status check (also related to load balancer).
Health endpoint must include statuses of underlying databases (mysql, postgres, sqlite) + redis - this depends which **database** provider use. It could contain other important information (maybe caching or other important information relevant to the service running).

### Suggestion

Example of `/health` endpoint (example uses **mysql** underlying database) - in `dependencies` - Cosmian can add additional important information related to Cosmian service running :
```
{
  "status": "UP",                           -> global status
  "dependencies": {
    "database": {                           -> underlying database postgres, mysql, sqlite or redis
      "name": "mysql"
      "status": "UP",
      "message": "latency: 3ms"
    },
    "cache": {                              -> additional important stuff when running Cosmian KMS service
      "status": "UP",
      "elements": 13
    }
  }
}
```

Global status is **UP** if all **dependencies** (statuses) are **UP**, otherwise is **DOWN** (service is down - crucial components for service failed).

Database health checks - query:
- postgres - `SELECT 1`
- mysql - `SELECT 1`
- sqlite - `SELECT 1`
- redis - send `PING` -> return `PONG`

Manuthor:
Closes #672

### DIFF
--- a/.github/scripts/docker-compose-with-load-balancer.yml
+++ b/.github/scripts/docker-compose-with-load-balancer.yml
@@ -1,0 +1,112 @@
+---
+# CI/dev helper: multiple KMS instances behind an Nginx load balancer,
+# sharing a single PostgreSQL database.
+#
+# Goal: verify the load balancer can serve `/health`.
+#
+# Usage:
+#   docker compose -f .github/scripts/docker-compose.yml up -d
+#   curl -fsS http://localhost:8080/health
+
+name: scripts
+
+services:
+  postgres:
+    image: postgres:16
+    environment:
+      POSTGRES_USER: kms
+      POSTGRES_PASSWORD: kms
+      POSTGRES_DB: kms
+    ports:
+      - 5432:5432
+    volumes:
+      - postgres-data:/var/lib/postgresql/data
+    healthcheck:
+      test: [CMD-SHELL, pg_isready -U kms -d kms]
+      interval: 5s
+      timeout: 5s
+      retries: 30
+
+  # KMS instances. We assume the server listens on HTTP 0.0.0.0:9998 by default.
+  # If your server is TLS-only in your deployment, you can still keep this as plain HTTP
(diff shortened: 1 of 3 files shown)


crate/server/src/routes/health.rs
```diff
@@ -0,0 +1,88 @@
...
+    let start = Instant::now();
+
+    let db_dep = health_dependencies(&kms).await;
+    let global_status = if db_dep.status == HealthStatus::Up {
+        HealthStatus::Up
+    } else {
+        HealthStatus::Down
+    };
+
+    Ok(actix_web::HttpResponse::Ok().json(HealthResponse {
```
nSaso: Http status code should return `200` OK if service global status is UP, otherwise http status should return `503` - Service Unavailable - when service is DOWN (on `HealthStatus::Down`)

Manuthor: What about this commit c8ec7bf04ca8b2ca8beab80470fcbd84fb3ea355?

PR #697, opened 2026-01-29
fix(ui): Azure BYOK export

Manuthor:
The KMIP GetAttributes operation do not retrieve `tags` except if request ask for them.
The goal of this PR is to add a new WASM function to explicitely request tags to be returned in UI.

### DIFF
--- a/crate/cli/src/tests/kms/azure/mod.rs
+++ b/crate/cli/src/tests/kms/azure/mod.rs
@@ -1,17 +1,55 @@
 use std::fs;
 
-use cosmian_kmip::kmip_2_1::{kmip_objects::ObjectType, kmip_types::KeyFormatType};
+use openssl::{
+    pkey::{PKey, Private, Public},
+    rsa::Rsa,
+};
 use tempfile::TempDir;
 use test_kms_server::start_default_test_kms_server;
 
 use crate::{
     actions::kms::{
         azure::byok::{ExportByokAction, ImportKekAction},
-        rsa::keys::create_key_pair::CreateKeyPairAction,
         symmetric::keys::create_key::CreateKeyAction,
     },
     error::{KmsCliError, result::KmsCliResult},
 };
+
+/// Generate RSA keypair using OpenSSL (random size from 2048, 3072, or 4096 bits).
+///
+/// This mirrors AWS KMS "get-parameters-for-import" wrapping key specs and keeps
+/// the test independent from KMS RSA key generation/export actions.
+fn generate_rsa_keypair() -> KmsCliResult<(PKey<Private>, PKey<Public>)> {
+    let key_sizes = [2048_u32, 3072_u32, 4096_u32];
+    // Avoid introducing new RNG deps in the CLI crate's dev-deps.
+    let bits = key_sizes[std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| {
+            let len_u32 = u32::try_from(key_sizes.len()).unwrap_or(1);
+            let idx_u32 = d.subsec_nanos() % len_u32;
+            usize::try_from(idx_u32).unwrap_or(0)
+        })
+        .unwrap_or(0)];
+
+    let rsa = Rsa::generate(bits)
+        .map_err(|e| KmsCliError::Default(format!("Failed to generate RSA key: {e}")))?;
+    let private_key = PKey::from_rsa(rsa.clone())
+        .map_err(|e| KmsCliError::Default(format!("Failed to build private key: {e}")))?;
+    let public_key = PKey::from_rsa(
+        Rsa::from_public_components(
+            rsa.n()
+                .to_owned()
+                .map_err(|e| KmsCliError::Default(format!("Failed to clone modulus: {e}")))?,
+            rsa.e()
+                .to_owned()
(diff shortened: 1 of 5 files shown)


crate/cli/src/tests/kms/azure/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+    };
+
+    let pem = cosmian_kms_client_utils::export_utils::der_to_pem(
+        der.as_slice(),
+        KeyFormatType::PKCS8,
+        ObjectType::PublicKey,
+    )
+    .map_err(|e| KmsCliError::Default(format!("DER to PEM conversion failed: {e}")))?;
+
+    fs::write(&kek_pem_path, pem.as_slice())?;
```
HatemMn: While this is technically 100% correct, I have preferd (aws PR) to use **openssl** for this OP to keep this test free from noise (ie: if there is an issue with export or generate RSA actions, this isn't supposed to bug)

if possible, it might be more convenient (for refactoring purposes) to use this

```rust
/// Generate RSA keypair using openssl (random size from 2048, 3072, or 4096 bits)
/// This replaces the AWS KMS keypair generation for testing purposes.
/// The aws API equivalent command (on cli) is:
/// ```sh
///  aws kms get-parameters-for-import \
/// --key-id <YOUR_EXTERNAL_KEY_ID> \
/// --wrapping-algorithm RSAES_OAEP_SHA_256 \
/// --wrapping-key-spec RSA_3072 # or RSA_2048 or RSA_4096
/// ```
pub(crate) fn generate_rsa_keypair()
-> Result<(PKey<Private>, PKey<Public>), Box<dyn std::error::Error>> {
    // Randomly select key size from AWS-supported sizes
    let key_sizes = [2048, 3072, 4096];
    let mut rng = CsRng::from_entropy();
    let bits = key_sizes[(rng.next_u32() as usize) % key_sizes.len()];

    let rsa = Rsa::generate(bits)?;

    let private_key = PKey::from_rsa(rsa.clone())?;
    let public_key = PKey::from_rsa(Rsa::from_public_components(
        rsa.n().to_owned()?,
        rsa.e().to_owned()?,
    )?)?;

    Ok((private_key, public_key))
}
```

if this is time consuming, please simply assign an issue to me and I will do it in here : https://github.com/Cosmian/kms/pull/681

crate/cli/src/tests/kms/azure/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+        Some(vec![format!("kid:{kid}")]),
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+    )
+    .map_err(|e| KmsCliError::Default(format!("Failed to build Locate request: {e}")))?;
```
HatemMn: the import action also tags "azure"

possible to add this to the test too ? It's the main UI detail that raised the original bug

crate/cli/src/tests/kms/azure/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+        kek_id: imported_kek_id,
+        byok_file: Some(byok_file.clone()),
+    }
+    .run(kms_client)
+    .await?;
+
+    // Assert byok file written
+    let contents = std::fs::read_to_string(&byok_file)?;
+    assert!(contents.contains("\"ciphertext\""));
+    assert!(contents.contains("\"kid\""));
```
HatemMn, line 105: "ideally", we would want to simulate what azure does by unwrapping the key

if you decided to follow my idea of using openssl on the other comment, we can do that here

https://github.com/Cosmian/kms/pull/681/changes#diff-20172d60dca3be22346346aa0f03e0cfaee47d3528523cec8f9d039a81372a5c

to not waste time here are the util functions I made for the matter, they can serte as model

We can also "report" this task for the aws PR

crate/wasm/src/wasm.rs
```diff
@@ -1670,6 +1670,32 @@ pub fn get_attributes_ttlv_request(unique_identifier: String) -> Result<JsValue,
...
+
+    let request = GetAttributes {
+        unique_identifier: Some(unique_identifier),
+        attribute_reference,
+    };
+
+    let objects = to_ttlv(&request).map_err(|e| JsValue::from(e.to_string()))?;
+    serde_wasm_bindgen::to_value(&objects).map_err(|e| JsValue::from(e.to_string()))
+}
+
```
HatemMn, line 1698: For a starter, I think this function's name should be `get_attributes_ttlv_request_with_tags` due to how it behaves ( **if nothing is changed** )

However after reading it's other counterpart : 
```rust
// Attributes request
#[wasm_bindgen]
pub fn get_attributes_ttlv_request(unique_identifier: String) -> Result<JsValue, JsValue> {
    let unique_identifier = UniqueIdentifier::TextString(unique_identifier);
    let request = GetAttributes {
        unique_identifier: Some(unique_identifier),
        attribute_reference: None,
    };
    let objects = to_ttlv(&request).map_err(|e| JsValue::from(e.to_string()))?;
    serde_wasm_bindgen::to_value(&objects).map_err(|e| JsValue::from(e.to_string()))
}
```


I rather consider this function should actually 'force' any other possible attribute not just tags

Due to certification & debug priority over this comment, do not hesitate to assign it to me on some issue with the other comments (AWS PR has direct relation with all of this)

PR #724, opened 2026-02-19
monitoring

p0wline:


monitoring/otel-collector-config.yaml
Manuthor, line 1: Please rename it with the prefix `ci-`

p0wline: Done

PR #737, opened 2026-02-26
chore: reafctoring + dedup of the UI codebase

HatemMn:
## Update : this also covers this PR : https://github.com/Cosmian/kms/pull/876

__________________________________

# Globally:
- [x] Reading some configurations that might be obsolete, deleting or moving them
- [x] Creating an AGENTS context file - symlinking all the other AI files to that one.

> **Purpose of this file**: This is the single source of truth for any AI agent
> (Copilot, Cursor, Cline, Claude Code, etc.) working on the Cosmian KMS codebase. It
> explains project structure, build commands, CI workflows, coding conventions,
> and troubleshooting steps so the agent can act autonomously and correctly.

- [x] clean root (whenever was possible)
- [x] structure ui/src/ folder
- [x] new folder called **actions** that contains basically what the ui does beneath the menu. There was not standard naming, I took the choice to name it "actions" because it's a term we often use in the Rust cli and most people will know what it means
- [x] adding [knip.dev](knip.dev) to the UI and the CI of the UI in order to avoid having dead code hanging in the future. It now runs on the CI thanks to the fix documented just below :

you can notice this on the **package.json:** 
```json
        "supportedArchitectures": {
            "os": ["current", "linux", "darwin", "win32"],
            "cpu": ["current", "x64", "arm64"]
        },
```
[Due to CI issues to compile the correct the wasm package](https://github.com/Cosmian/kms/actions/runs/23903605921/job/69706380843?pr=737), I simply added this line that forces every dev to install a few extra small .node files for other platforms than his own.

---> The benefit: the lockfile always has every binding regardless of who generates it ( CI on linux, mac dev, windows dev, and if a system is missing just add it to that list )

# ⚠️ PLEASE NOTE

If you are using an AI coding assistant that does not natively support [AGENTS.md](AGENTS.md) (e.g. Claude Code, at the time of
writing), create a symbolic link so it picks up the agent instructions automatically:

```sh
ln -s AGENTS.md CLAUDE.md
```

### DIFF
--- a/nix/expected-hashes/cli.vendor.dynamic.darwin.sha256
+++ b/nix/expected-hashes/cli.vendor.dynamic.darwin.sha256
@@ -1,1 +1,1 @@
-sha256-I1lt1wmibJbq16NO4MwA7NU55Ex59stxpyCS199HbWE=
+sha256-O0C/pmXwPx3OK2hvFe3+lHehTg7eDbBbNMvLLtsNvPw=
--- a/nix/expected-hashes/cli.vendor.static.darwin.sha256
+++ b/nix/expected-hashes/cli.vendor.static.darwin.sha256
@@ -1,1 +1,1 @@
-sha256-An0USOAShNjnqWD21trgpkMxk20z4drwmT1Ly7tTfZY=
+sha256-Af5T4+FAci+cUFAys5P4AmeL5G9cq9GumvFKYFWzL1c=


_typos.toml
Manuthor, line 1: I need this one!

HatemMn: I will address this, thanks

Issue #734, opened 2026-02-23
[UI Enhancement] Allow multiple value selection for column filters in the results table


chokoblitz:
### Summary

Currently, the **Locate Cryptographic Objects** results table displays objects with columns 
such as **Type**, **Key Format Type**, and **State**, but there is no way to filter 
the results directly from the table. Adding inline column filters with **multi-value 
selection** would allow users to quickly narrow down their view without running a new search.

### Current Behavior

- The search form allows pre-filtering before the query is executed ✅
- The results table has no inline filtering capability ❌
- To see e.g. only `PrivateKey` and `SymmetricKey` objects, the user must run two 
  separate searches

### Expected Behavior

Each column in the results table should support an inline filter with **multi-value 
selection**, for example:

- **Type**: `[ ] PublicKey` `[x] PrivateKey` `[x] SymmetricKey`
- **Key Format Type**: `[x] TransparentECPrivateKey` `[ ] Raw` ...
- **State**: `[x] Active` `[x] Deactivated`

Selected values would be combined with **OR logic** within the same column 
(e.g. `Type = PrivateKey OR SymmetricKey`), and **AND logic** across columns 
(e.g. `Type = PrivateKey AND State = Active`).

### Environment

- Cosmian KMS version: `5.16.2`
- UI: KMS Web Interface

Issue #732, opened 2026-02-23
[UI Enhancement] Add sort order (asc/desc) by column in the "Locate Cryptographic Objects" view


chokoblitz:
### Summary

The **Locate Cryptographic Objects** view already provides a rich set of search/filter criteria 
(Tags, Cryptographic Algorithm, Cryptographic Length, State, Object Type, Key Format Type, 
Linked Objects). However, the **results table** does not allow sorting by any column. 
Adding ascending/descending sort on the result columns would complement the existing 
filtering capabilities and greatly improve usability.

### Current Behavior

- The search form allows filtering objects by multiple criteria ✅
- The result list is paginated (10/page) ✅
- **No sort control is available on the result table columns** ❌

### Expected Behavior

Each column header in the results table should be clickable to toggle sort order:
- **Type** (PublicKey, PrivateKey, SymmetricKey, ...) – categorical sort
- **Key Format Type** (TransparentECPublicKey, Raw, ...) – categorical sort
- **State** (Active, Deactivated, ...) – categorical sort

Sort direction should cycle through: `Ascending → Descending → Default`.

### Environment

- Cosmian KMS version: `x.x.x`
- UI: KMS Web Interface

PR #738, opened 2026-02-26
feat:  Multiple UI Fixes

HatemMn:
# Summary

c53978b	feature ->	add prettier and format everything
28c641c	feature ->	sorting located keys / fixes https://github.com/Cosmian/kms/issues/732
38285bd	fix ->	update hashes and fix stuff
a7b3e12	feature ->	add filters to the locate page / fixes https://github.com/Cosmian/kms/issues/734
da07fce	feature ->	add WASM error message -> https://github.com/Cosmian/kms/issues/764

### DIFF
--- a/ui/src/Locate.tsx
+++ b/ui/src/Locate.tsx
@@ -552,32 +552,8 @@ const LocateForm: React.FC = () => {
                             setRes(`${filtered.length} Object(s) located.`);
                             return;
                         }
-                        if (keyFormatType) {
-                            const targetKft = kftNorm(keyFormatType);
-                            merged = merged.filter((r) => {
-                                const v = r.meta?.["key_format_type"] as string | undefined;
-                                return v ? kftNorm(v) === targetKft : false;
-                            });
-                        }
-
-                        setObjects(merged);
-                        setRes(`${merged.length} Object(s) located.`);
-                    } catch {
-                        let filtered = enriched;
-                        if (stateVal) {
-                            const target = normalizeState(stateVal);
-                            filtered = filtered.filter((r) => normalizeState(r.state) === target);
-                        }
-                        if (keyFormatType) {
-                            const targetKft = kftNorm(keyFormatType);
-                            filtered = filtered.filter((r) => {
-                                const v = r.meta?.["key_format_type"] as string | undefined;
-                                return v ? kftNorm(v) === targetKft : false;
-                            });
-                        }
-
-                        setObjects(filtered);
-                        setRes(`${filtered.length} Object(s) located.`);
+                    } catch (e) {
+                        console.error("Fallback Locate without KFT failed:", e);
                     }
                 }
                 // Still nothing: show explicit 0 objects
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -5,12 +5,12 @@ settings:
     excludeLinksFromLockfile: false
 
 overrides:
-  esbuild: ^0.25.10
-  '@babel/runtime': ^7.26.10
-  minimatch: '>=10.2.1'
-  rollup: '>=4.59.0'
-  ajv: ^6.14.0
-  flatted: '>=3.4.2'
+    esbuild: ^0.25.10
+    "@babel/runtime": ^7.26.10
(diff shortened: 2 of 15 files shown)


ui/src/Locate.tsx
```diff
@@ -756,92 +761,81 @@ const LocateForm: React.FC = () => {
...
-                                );
-                            })()}
+                            <Table<LocateObjectRow>
+                                dataSource={objects || []}
+                                rowKey="object_id"
+                                pagination={{
+                                    pageSize: 10,
+                                    showSizeChanger: true,
+                                    pageSizeOptions: [10, 20, 50, 100],
+                                }}
```
HatemMn, line 779: Nothing much changed from the previous version but in a nushell this diff looks huge because instead of manual state calculus like
```
    const [currentPage, setCurrentPage] = useState<number>(1);
    const [pageSize, setPageSize] = useState<number>(10);
```

Now antd handles everything internally (result is the same)

HatemMn (commented):
Had to take out prettier of the CI because nix does not support it

This is still the relevant commit : https://github.com/Cosmian/kms/pull/738/changes/4d29bb3c19b86b2173cdd9f3b26f23b133bb20cd

.github/scripts/test_wasm.sh
```diff
@@ -200,8 +200,10 @@ if [ -f ui/pnpm-lock.yaml ]; then
   fi
   run_ui pnpm install --frozen-lockfile
   run_ui pnpm run lint
+  run_ui npm run check
```
charming-wicket-5502: `npm run check` runs eslint but the line above `pnpm run lint` already runs eslint --fix

ui/src/Locate.tsx
```diff
@@ -742,80 +769,107 @@ const LocateForm: React.FC = () => {
...
+                                            key: "attributes.ObjectType",
+                                            sorter: (a: LocateObjectRow, b: LocateObjectRow) =>
+                                                (a.attributes?.ObjectType ?? "").localeCompare(b.attributes?.ObjectType ?? ""),
+                                            filters: [
+                                                ...objectTypes.map((t) => ({ text: t.label, value: t.value })),
+                                                { text: "N/A", value: "N/A" },
+                                            ],
+                                            // OpaqueObject among probably others are not keys and have no KeyFormatType so N/A is a catch-all handled separately
+                                            onFilter: (value: React.Key | boolean, record: LocateObjectRow) => {
+                                                if (value === "N/A") return !record.attributes?.ObjectType;
```
charming-wicket-5502: Could not find similar edge case management in antd doc exemple, is this because ObjectType contains a "N/A" value ? Would it be preferable to remove it from the objectTypes list locally to not display the option at all ?

.github/scripts/test_wasm.sh
```diff
@@ -200,8 +200,10 @@ if [ -f ui/pnpm-lock.yaml ]; then
...
   run_ui pnpm run test:unit
   run_ui pnpm audit --audit-level high
```
HatemMn, line 192: test

.github/scripts/test_wasm.sh
HatemMn: True,

Deleted run_ui pnpm run lint

it was probably a mistake to put it there, it does nothing

ui/src/Locate.tsx
HatemMn: image

This is a screenshot taken after removing the line

```tsx
if (value === "N/A") return !record.attributes?.ObjectType;
```

>  is this because ObjectType contains a "N/A" value ? 

Exactly !

> Could not find similar edge case management in antd doc exemple

Actually, I don't think the apparition of that term is related to antd but rather to some WASM function that would be very painful to "find" - to my understanding and I might be partially wrong some WASM related code uses "N/A" when he does not find an object type to write

> Would it be preferable to remove it from the objectTypes list locally to not display the option at all ?

This isn't a bug, but rather a poorly documented manner of handling the absence of a type. I am against not displaying it and prefer gracefully handling it instead

My method is one possible way of handling those "N/A" values, don't hesitate to propose other ways if you can think of them

PR #771, opened 2026-03-09
fix: fail KMIP operation when multiple keys are found to process the operation

### LINKED ISSUE #731: [Server] Encrypt and Sign KMIP operations
Verify than both KMIP operations fail when multiple keys are found at the beginning of operation.

Manuthor:
Closes #731

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,13 +45,17 @@ in this repository under `crate/clients/ckms/`:
   system pnpm and the nix-shell pnpm
 - add `--header`/`-H` flag and `custom_headers` config option to forward arbitrary HTTP headers with every request, enabling use behind zero-trust proxies such as Cloudflare Access ([#138](https://github.com/Cosmian/cli/issues/138))
 
+#### HSM signing via Crypto Oracles
+
+- KMIP `Sign` operation is now delegated to HSM Crypto Oracles via PKCS#11 `C_SignInit`/`C_Sign` ([#771](https://github.com/Cosmian/kms/pull/771))
+
 #### White labeling
 
 - The vendor identification string used in KMIP `VendorAttribute` operations is now
   configurable via `--vendor-identification` (env: `KMS_VENDOR_IDENTIFICATION`, default:
   `"cosmian"`); reported back by `QueryServerInformation` responses
 - The WASM module exposes `set_vendor_id(id)` and `query_server_information_ttlv_request()`
-  so the UI can synchronise its vendor ID with the server at startup
+  so the UI can synchronize its vendor ID with the server at startup
 - New `loginCardColor` field in `branding.json` to control the login card background color
 - New blank starter theme at `ui/public/themes/blank/` with SVG placeholder assets
 
@@ -63,6 +67,7 @@ in this repository under `crate/clients/ckms/`:
   `/usr/local/cosmian/` on uninstall
 - **macOS build**: retry loop in `nix/scripts/package_dmg.sh` handles intermittent
   `hdiutil: create failed - Resource busy` CI errors
+- KMIP `Encrypt` and `Sign` now fail with a clear error when multiple eligible keys match the identifier, preventing silent key substitution ([#771](https://github.com/Cosmian/kms/pull/771))
 
 ### ⚙️ Build
 


crate/hsm/base_hsm/src/session/session_impl.rs
```diff
@@ -1253,6 +1276,72 @@ impl Session {
...
+            "Failed to sign data",
+            C_Sign,
+            self.handle,
+            data.as_mut_ptr(),
+            CK_ULONG::try_from(data.len())?,
+            signature.as_mut_ptr(),
+            &raw mut signature_len
+        );
+
+        signature.truncate(usize::try_from(signature_len)?);
```
tbrezot: Is this a normal behavior to receive two different `signature_length`s from the HSM?

crate/server/src/core/operations/encrypt.rs
```diff
@@ -112,59 +106,47 @@ pub(crate) async fn encrypt(kms: &KMS, request: Encrypt, user: &str) -> KResult<
...
+    // Phase 2 — Standard database path via shared selection function.
+    let mut owm = select_unique_key_for_operation(
+        "Encrypt",
+        &uids,
+        unique_identifier,
+        &[KmipOperation::Encrypt, KmipOperation::Get],
+        kms,
+        user,
+        |owm| {
+            // TODO check why usage masks are not checked for certificates
```
tbrezot: TODO ? (best not to keep any in the code for the certification)

crate/server/src/core/operations/sign.rs
```diff
@@ -51,17 +55,62 @@ pub(crate) async fn sign(kms: &KMS, request: Sign, user: &str) -> KResult<SignRe
...
+    // Phase 1 — Oracle (prefix) UIDs.
+    for uid in &uids {
+        if let Some(prefix) = has_prefix(uid) {
+            if !kms.database.is_object_owned_by(uid, user).await? {
+                let ops = kms
+                    .database
+                    .list_user_operations_on_object(uid, user, false)
+                    .await?;
+                if !ops.iter().any(|p| *p == KmipOperation::Sign) {
+                    continue;
```
tbrezot: There should be a function for doing this, no ? the one which gives the rights if you have `Get` (which is not the case here btw).

If we want to reduce the surface for bugs, we should strive to reuse key functions for sensitive operations like this one.

crate/server/src/core/operations/state_utils.rs
```diff
@@ -51,6 +61,96 @@ pub(crate) fn get_effective_state(owm: &ObjectWithMetadata) -> KResult<State> {
     Ok(State::PreActive)
 }
 
+/// Select exactly one key from a set of candidate UIDs for a cryptographic operation.
+///
+/// `candidate_uids` is a `HashSet` as returned by `uid_utils::uids_from_unique_identifier`.
```
tbrezot, line 90: Maybe internalize this call?

crate/hsm/base_hsm/src/session/session_impl.rs
Manuthor: You don't receive two different lengths. This is the standard PKCS#11 two-pass idiom:

- First call (pSignature = NULL): asks the HSM "how many bytes will the signature be?" — the HSM writes the maximum buffer size needed into [signature_len](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
- Second call (pSignature = buffer): actually signs the data. The HSM writes the signature into the buffer and updates [signature_len](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to the actual number of bytes written.

crate/server/src/core/operations/encrypt.rs
Manuthor: Correct and done in 6f101c95

crate/server/src/core/operations/sign.rs
Manuthor: Done in 6f101c95

crate/hsm/base_hsm/src/session/session_impl.rs
tbrezot: My point was: shouldn't this `truncate` be replaced by a check that the second length is the same as the first one?

Manuthor: done in 42d16756

crate/server/src/core/operations/state_utils.rs
Manuthor: done in 42d16756

PR #787, opened 2026-03-13
feat: support ML-KEM, hybrid KEM, ML-DSA and SLH-DSA

Manuthor:
#### PQC UI Enhancements & Hash Operation

- **PQC pages generalized**: Encapsulate/Decapsulate pages now show "PQC KEM" (covering ML-KEM
  and Hybrid KEM: X25519MLKEM768, X448MLKEM1024); Sign/Verify pages now show "PQC Signature"
  (covering ML-DSA and SLH-DSA). Key creation descriptions updated to list all four PQC families.
- **Hash operation page**: new "Hash" page under the Symmetric menu computes server-side
  cryptographic hashes (SHA-256, SHA-384, SHA-512, SHA3-224, SHA3-256, SHA3-384, SHA3-512)
  via the KMIP Hash operation. Supports file upload or text input.
- **WASM hash bindings**: `hash_ttlv_request()`, `parse_hash_ttlv_response()`, and
  `get_hash_algorithms()` added to the WASM module.
- **E2E tests**: Playwright roundtrip tests added for Hybrid KEM (X25519MLKEM768
  encapsulate/decapsulate) and SLH-DSA (SLH-DSA-SHA2-128s sign/verify).

#### Post-Quantum Cryptography (ML-KEM + ML-DSA)

Full support for NIST post-quantum algorithms via OpenSSL 3.x default provider
(non-FIPS builds only):

- **ML-KEM** (Key Encapsulation Mechanism): ML-KEM-512, ML-KEM-768, ML-KEM-1024 — key pair
  creation, encapsulation, and decapsulation via KMIP Encrypt/Decrypt operations
- **ML-DSA** (Digital Signature Algorithm): ML-DSA-44, ML-DSA-65, ML-DSA-87 — key pair
  creation, signing, and verification via KMIP Sign/SignatureVerify operations
- New KMIP enumeration values for all six PQC algorithms
- Server dispatch for PQC key creation, encrypt/decrypt (KEM), and sign/verify
- CLI actions: `ckms pqc keys create`, `ckms pqc encapsulate`, `ckms pqc decapsulate`,
  `ckms pqc sign`, `ckms pqc verify`
- WASM bindings: `create_pqc_key_pair_ttlv_request()`, `get_pqc_algorithms()`
- Web UI pages: PQC key creation, ML-KEM encapsulate/decapsulate, ML-DSA sign/verify
- Playwright E2E tests for all PQC UI flows
- CLI integration tests for ML-KEM and ML-DSA roundtrips

#### Configurable Hybrid KEM merged into PQC

- Merged the standalone `ckms kem` subcommand into `ckms pqc` — the four hybridized KEM
  algorithms (ml-kem-512-p256, ml-kem-768-p256, ml-kem-512-curve25519, ml-kem-768-curve25519)
  are now created, encapsulated, and decapsulated through the standard PQC workflow
- Auto-detection in encapsulate response handles both PQC and ConfigurableKEM response formats
- WASM bindings updated with the 4 hybrid algorithms
- UI branding supports `hiddenPqcAlgorithms` to hide specific algorithms from the PQC dropdown
- CLI and ckms integration tests added for configurable hybrid KEM roundtrips

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,179 +1,347 @@
-# Cosmian KMS — AI Agent Guide
+# Cosmian KMS — AI Agent Instructions
 
-Cosmian KMS is a high-performance, open-source FIPS 140-3 compliant Key Management System written in Rust.
-It implements KMIP 2.1 over HTTP/TLS and supports AES, RSA, EC, ML-KEM, ML-DSA, SLH-DSA, Covercrypt, and more.
+> **Purpose of this file**: This is the single source of truth for any AI agent
+> (Copilot, Cursor, Cline, etc.) working on the Cosmian KMS codebase. It
+> explains project structure, build commands, CI workflows, coding conventions,
+> and troubleshooting steps so the agent can act autonomously and correctly.
 
-## Build & Test cheatsheet
+Cosmian KMS is a high-performance, open-source **FIPS 140-3** compliant Key
+Management System written in **Rust**. It implements **KMIP 2.1** over HTTP/TLS
+and supports AES, RSA, EC, ML-KEM, ML-DSA, SLH-DSA, Covercrypt, and more.
+
+---
+
+## 1. Build & test cheatsheet
 
 ```bash
-# Build (FIPS mode is the default)
-cargo build
-cargo build --features non-fips   # non-FIPS: extra algorithms, legacy provider, PQC
+# ── Build ────────────────────────────────────────────────────────────────
+cargo build                          # FIPS mode (default)
+cargo build --features non-fips      # non-FIPS: extra algorithms, PQC, Covercrypt
 
-# Test
-cargo test-fips                    # alias: test --lib --workspace
-cargo test-non-fips                # alias: test --lib --workspace --features non-fips
-cargo test -p cosmian_kms_server   # single package
+# ── Test (cargo aliases defined in .cargo/config.toml) ───────────────────
+cargo test-fips                      # test --lib --workspace
+cargo test-non-fips                  # test --lib --workspace --features non-fips
+cargo test -p cosmian_kms_server     # single crate
 cargo test -p cosmian_kms_cli
 
-# Lint
-cargo clippy-all                   # alias: clippy --workspace --all-targets --all-features -- -D warnings
-cargo format                       # alias: fmt --all -- --check
+# ── Lint ─────────────────────────────────────────────────────────────────
+cargo clippy-all                     # clippy --workspace --all-targets --all-features -- -D warnings
+cargo format                         # fmt --all -- --check
 
-# Run
+# ── Run locally ──────────────────────────────────────────────────────────
 cargo run --bin cosmian_kms -- --database-type sqlite --sqlite-path /tmp/kms-data
 
-# Probe (expect a KMIP validation error, not a 404)
+# ── Smoke-test (expect 422, not 404) ────────────────────────────────────
 curl -s -X POST -H "Content-Type: application/json" -d '{}' http://localhost:9998/kmip/2_1
 ```
 
-DB test environment variables (start backends with `docker compose up -d`):
+### Cargo aliases (`.cargo/config.toml`)
 
-- `KMS_POSTGRES_URL=postgresql://kms:kms@127.0.0.1:5432/kms`
-- `KMS_MYSQL_URL=mysql://kms:kms@localhost:3306/kms`
-- `KMS_SQLITE_PATH=data/shared`
+| Alias | Expands to |
+|---|---|
+| `format` | `fmt --all -- --check` |
+| `build-all` | `build --workspace --all-targets --all-features --bins` |
+| `test-fips` | `test --lib --workspace` |
+| `test-non-fips` | `test --lib --workspace --features non-fips` |
+| `clippy-all` | `clippy --workspace --all-targets --all-features -- -D warnings` |
 
-Notes:
+### Database test environment
 
-- MySQL tests are currently disabled in CI
-- Redis-findex tests are skipped in FIPS mode
+Start backends with `docker compose up -d`, then set:
 
-## Workspace layout
+| Variable | Value |
+|---|---|
+| `KMS_POSTGRES_URL` | `postgresql://kms:kms@127.0.0.1:5432/kms` |
+| `KMS_MYSQL_URL` | `mysql://kms:kms@localhost:3306/kms` |
+| `KMS_SQLITE_PATH` | `data/shared` |
+
+> MySQL tests are currently disabled in CI.
+> Redis-findex tests are skipped in FIPS mode.
+
+---
+
+## 2. Workspace layout
 
 ```text
 crate/
-  access/           cosmian_kms_access        — access-control utilities
+  access/           cosmian_kms_access         — access-control utilities
   cli/              cosmian_kms_cli            — CLI client binary
   clients/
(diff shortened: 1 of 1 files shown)


crate/cli/src/actions/kms/pqc/keys/create_key_pair.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub(crate) const fn to_cryptographic_algorithm(self) -> CryptographicAlgorithm {
+        match self {
+            Self::MlKem512 => CryptographicAlgorithm::MLKEM_512,
+            Self::MlKem768 => CryptographicAlgorithm::MLKEM_768,
+            Self::MlKem1024 => CryptographicAlgorithm::MLKEM_1024,
+            Self::MlDsa44 => CryptographicAlgorithm::MLDSA_44,
+            Self::MlDsa65 => CryptographicAlgorithm::MLDSA_65,
+            Self::MlDsa87 => CryptographicAlgorithm::MLDSA_87,
+            Self::X25519MlKem768 => CryptographicAlgorithm::X25519MLKEM768,
+            Self::X448MlKem1024 => CryptographicAlgorithm::X448MLKEM1024,
```
tbrezot, line 106: why not implement those in the configurable KEM?

crate/crypto/src/crypto/pqc/ml_kem.rs
```diff
@@ -0,0 +1,295 @@
...
+                "ML-KEM encapsulate: EVP_PKEY_CTX_new failed".to_owned(),
+            ));
+        }
+        let _ctx_guard = CtxGuard(ctx);
+
+        if openssl_sys::EVP_PKEY_encapsulate_init(ctx, ptr::null()) != 1 {
+            return Err(CryptoError::Default(
+                "ML-KEM encapsulate: EVP_PKEY_encapsulate_init failed".to_owned(),
+            ));
+        }
```
tbrezot: Isn't there a memory leak here in case of an error because the the ciphertext is never freed?

crate/crypto/src/crypto/pqc/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+        if pkey.is_null() {
+            return Err(CryptoError::Default(format!(
+                "EVP_PKEY_Q_keygen failed for {algorithm_name}: {}",
+                openssl::error::ErrorStack::get()
+            )));
+        }
+
+        let bits = u32::try_from(openssl_sys::EVP_PKEY_bits(pkey)).map_err(|e| {
+            CryptoError::Default(format!("EVP_PKEY_bits returned negative value: {e}"))
+        })?;
```
tbrezot: Memory leak when an error is returned as `pkey` is not freed. Same below.

crate/crypto/src/crypto/pqc/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+            pkey,
+            buf.as_mut_ptr(),
+            ptr::from_mut(&mut len),
+        ) != 1
+        {
+            return Err(CryptoError::Default(
+                "EVP_PKEY_get_raw_private_key failed".to_owned(),
+            ));
+        }
+        buf.truncate(len);
```
tbrezot: What if for some reason not all bytes could be written ? Is this a bug ?

crate/crypto/src/crypto/pqc/ml_kem.rs
```diff
@@ -0,0 +1,295 @@
...
+            return Err(CryptoError::Default("ML-KEM decapsulate failed".to_owned()));
+        }
+
+        shared_secret.truncate(secret_len);
+        Ok(shared_secret)
+    }
+}
+
+/// RAII guard to free an `EVP_PKEY_CTX`.
+pub(crate) struct CtxGuard(pub(crate) *mut openssl_sys::EVP_PKEY_CTX);
```
tbrezot, line 247: This technique may be used everywhere in the new bindings (e.g. for pkeys (btw why can't you reuse the `PKey` from the openssl crate?: 
```
generic_foreign_type_and_impl_send_sync! {
    type CType = ffi::EVP_PKEY;
    fn drop = ffi::EVP_PKEY_free;

    /// A public or private key.
    pub struct PKey<T>;
    /// Reference to `PKey`.
    pub struct PKeyRef<T>;
}
```
which seems to implement the drop with the correct free function))

crate/crypto/src/crypto/pqc/ml_kem.rs
```diff
@@ -0,0 +1,295 @@
...
+            ptr::from_mut(&mut wrapped_len),
+            shared_secret.as_mut_ptr(),
+            ptr::from_mut(&mut secret_len),
+        ) != 1
+        {
+            return Err(CryptoError::Default("ML-KEM encapsulate failed".to_owned()));
+        }
+
+        wrapped_key.truncate(wrapped_len);
+        shared_secret.truncate(secret_len);
```
tbrezot: same here, is this a bug if `wrapped_key.len() != wrapped_len` ? same for the shared secret

tbrezot (commented):
Potential memory leaks in the ML-KEM bindings. Maybe try reusing more objects from the OpennSSL crate?

crate/crypto/src/crypto/pqc/mod.rs
Manuthor: Fixed: wrapped raw `EVP_PKEY` pointers in a `PKeyGuard` RAII struct (analogous to `CtxGuard` already in `ml_kem.rs`) that calls `EVP_PKEY_free` on drop. All error paths via `?` now correctly free the key. Note: using `PKey<T>` from the openssl crate directly would require importing the `ForeignType` trait from `foreign_types_shared`, which is not a direct workspace dependency, so a lightweight local wrapper was added instead.

crate/crypto/src/crypto/pqc/mod.rs
Manuthor: Fixed: added explicit size-mismatch checks in `evp_pkey_get_raw_private` and `evp_pkey_get_raw_public` — if the second call returns a different length than the first (size-query) call, a `CryptoError` is returned instead of silently truncating.

crate/crypto/src/crypto/pqc/ml_kem.rs
Manuthor: Adopted: all raw `EVP_PKEY` pointers are now wrapped in `PKeyGuard` (a `CtxGuard`-style RAII struct) — see reply on the memory-leak comment for why we chose a local wrapper over `PKey<T>` from the openssl crate.

crate/crypto/src/crypto/pqc/ml_kem.rs
Manuthor: Fixed: replaced the silent `truncate` calls in `ml_kem_encapsulate_raw`, `ml_kem_decapsulate_raw`, `encapsulate_raw` (hybrid), and `decapsulate_raw` (hybrid) with explicit size checks — a mismatch between the preallocated size and the size returned by OpenSSL after the actual operation now returns a hard `CryptoError`.

CLAUDE.md
HatemMn, line 1: Is it possible to keep this file for this PR: https://github.com/Cosmian/kms/pull/737 and use it only locally for now ?

AGENTS.md
HatemMn, line 1: Same as Claude md, possible to put this here : https://github.com/Cosmian/kms/pull/737 if it was written explicitly ?

PR #801, opened 2026-03-18
fix: multiple hsm fixes

### LINKED ISSUE #745: key_encryption_key use CKA_LABEL for key lookup which is not unique
In order to force all keys to be wrapped by an HSM key, I've set up a key_encryption_key, fallowing this documentation:  

https://docs.cosmian.com/key_management_system/hsms/hsm_operations/#creating-a-kms-key-wrapped-by-an-hsm-key 

My kms.toml content is : 

```
# Force all keys to be wrapped by an HSM key
key_encryption_key = "hsm::4::master_kek"
```
The PKCS#11 standard does not require CKA_LABEL to be unique.
You can create as many keys as you want with the same label, I've created many master_kek keys : 

```
$ pkcs11-tool  --slot 1 --list-objects | grep ' master_kek' | wc -l
42
```
Expected results : 

- At least the documentation should alert the user of this issue ;
- The KSM should not start if more than one _master_kek_ is present ;
- CKA_UNIQUE_ID should be used from PKCS#11 3.0 (cf : https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html#_Toc29976567 ) ;

### LINKED ISSUE #761: key_encryption_key ownership sharing for keys wraping
As you wrote in the KMS documentation : 

> The KMS server can automatically wrap all KMS keys with a specific HSM key. This is done by setting the key_encryption_key property in the TOML server configuration file or using the corresponding command line switch.

Source : https://docs.cosmian.com/key_management_system/hsms/hsm_operations/

I have been experiencing a significant limitation with that feature: utilizing the KMS is restricted to the client whose certificate correspond to hsm_admin.

When the KMS retrieve an key from the HSM store : https://github.com/Cosmian/kms/blob/develop/crate/interfaces/src/hsm/hsm_store.rs#L101

You check the ownership on the key. But the key_encryption_key **is share by all** the KMS users.

When the common name (CN) of the client does not match the hsm_admin object wrapping is **denied** with the fallowing message : 

`2026-03-03T13:39:58.866690Z  WARN ThreadId(11) cosmian_kms_server::routes: crate/server/src/routes/mod.rs:57: [error_response] 422 Unprocessable Entity - Not Supported: The user client does not have the permission to wrap using the key hsm::1::master_kek`

In that case, my client CN was client and hsm_admin was set to admin.

In order to wrap all KMS keys : client CN must equal hsm_admin.

It means that only one client car wrap keys in this setup.

Best Regards,

### LINKED ISSUE #762: Wrapping key is sensitive and cannot be exported from the HSM.
Hi,

I try to unwrap a key from the KMS using the HSM key, and I've got the fallowing message : 

```
$ cosmian kms sym keys unwrap test.json -i hsm::1::master_kek

ERROR: Get: REST Request Failed: /kmip/2_1: Database Error: Ext. store error: This key is sensitive and cannot be exported from the HSM.
```

This key master key was created with p11tool with the fallowing command : 

$ pkcs11-tool   --slot 1  --key-type AES:16 --keygen --label master_kek

And the test.json content is a simple symetic key created like : 
```
$  cosmian kms sym keys create --algorithm aes --number-of-bits 256 --wrapping-key-id hsm::1::master_kek
The symmetric key was successfully generated.
          Unique identifier: 5f386ce8-9490-44d3-afa3-8df0b13ca6e9
$ cosmian kms sym keys export test.json --key-id 5f386ce8-9490-44d3-afa3-8df0b13ca6e9
```

Manuthor:
#### HSM multi-admin support with wildcard

`hsm_admin` is now a list of KMS usernames with HSM admin privileges. Use `["*"]` to grant all
authenticated users access to all HSM operations:
  - TOML: `hsm_admin = ["alice", "bob"]`;
  - CLI: `--hsm-admin alice --hsm-admin bob`; 
  - env: `KMS_HSM_ADMIN=alice,bob`.

### 🐛 Bug Fixes

- **HSM: CKA_ID missing on HSM-created keys**: Keys generated via the HSM PKCS#11 path were stored
  without a `CKA_ID`, making them invisible to some PKCS#11 tools. The KMS now sets `CKA_ID` at
  key creation time for all HSM backends (Proteccio, Utimaco, SoftHSM2).
- **HSM: spurious `CKA_VERIFY_RECOVER` warning from pkcs11-tool**: `pkcs11-tool --list-objects`
  emitted `CKR_ATTRIBUTE_TYPE_INVALID` for `CKA_VERIFY_RECOVER` on RSA public keys. The KMS now
  calls `C_SetAttributeValue(CKA_VERIFY_RECOVER=false)` after key generation (best-effort, silently
  ignored by HSMs that do not implement the attribute).

Closes #761
Closes #762
Closes #745

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -82,7 +82,7 @@ Expected response is a KMIP validation error, confirming the server is alive.
 
 ## Update CHANGELOG.md
 
-For each changes, compare the current git-branch with the upstream branch and summarize it on `CHANGELOG.md` on 1 line. This helps maintain a clear history of changes for users and contributors. Take example from last entries of `CHANGELOG.md` for formatting and style.
+For each changes, compare the current git-branch with the upstream branch and summarize it on `CHANGELOG.md` on 1 line in the top section of the file (next release). This helps maintain a clear history of changes for users and contributors. Take example from last entries of `CHANGELOG.md` for formatting and style.
 
 ## Repository layout (high level)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,10 +47,16 @@ to the test matrix so regressions are caught automatically:
 - **HSM: CKA_ID missing on HSM-created keys**: Keys generated via the HSM PKCS#11 path were stored
   without a `CKA_ID`, making them invisible to some PKCS#11 tools. The KMS now sets `CKA_ID` at
   key creation time for all HSM backends (Proteccio, Utimaco, SoftHSM2).
-- **HSM: spurious `CKA_VERIFY_RECOVER` warning from pkcs11-tool**: `pkcs11-tool --list-objects`
-  emitted `CKR_ATTRIBUTE_TYPE_INVALID` for `CKA_VERIFY_RECOVER` on RSA public keys. The KMS now
-  calls `C_SetAttributeValue(CKA_VERIFY_RECOVER=false)` after key generation (best-effort, silently
-  ignored by HSMs that do not implement the attribute).
+- **HSM**: HSM key lookup (`get_object_handle`) now searches by `CKA_ID` first (primary
+  path for KMS-created keys) and falls back to `CKA_LABEL` for externally provisioned keys
+  that may not have `CKA_ID` set; `get_object_id` follows the same order
+- **HSM**: Non-admin users can now create KMS keys wrapped by the server-level
+  `key_encryption_key`; the ownership check is skipped for this shared server resource
+  ([#761](https://github.com/Cosmian/kms/issues/761))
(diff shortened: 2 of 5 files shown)


crate/server/src/config/command_line/hsm_config.rs
serene-kitfisto-8899, line 1: maybe using % asan wildcard for username will be more appropriate like mysql does : 
https://dev.mysql.com/doc/refman/8.4/en/account-names.html

Manuthor: Thanks for the link. I would rather keep the '*' character, being IMHO clearer from scratch.

PR #813, opened 2026-03-20
changed disabled flag for menu items to only work in production envir…

charming-wicket-5502:
…onements


HatemMn:
I just received a notification to review this like an hour ago, which is weird

Is this PR still up to date ? If I remember right, what the title indicate is done on another pr no ?

PR #835, opened 2026-03-27
Add OpenSSH support

### LINKED ISSUE #688: [Documention] RSA_AES_KEY_WRAP_SHA_256 wrapping/unwraping (2 steps)
**This isn't a bug - rather a request reminder for documentation**

While the KMS is able to do this operation, it lacks documentation on the matter

Docs can be added here : https://docs.cosmian.com/key_management_system/kmip/_export/

To acheive that hybird unwrapping scheme, we should provide : 

```rust
CryptographicParameters {
                cryptographic_algorithm: Some(CryptographicAlgorithm::RSA),
                // Note: We use "None" padding to route toward RSA AES Key Wrap, this is not a mistake
                // see: crate/crypto/src/crypto/wrap/unwrap_key.rs line 365
                padding_method: Some(PaddingMethod::None),
                hashing_algorithm: Some(HashingAlgorithm::SHA1),
                ..CryptographicParameters::default()
            },
```

Since counter-intuitive, this behavior is better to be documented (don't take this issue, I will do it. Ping me if something is not clear)

### LINKED ISSUE #695: kms --hsm-password shloud not be visible in ps auxww output
Hi,

I think that the content of the hsm password argument should not be visible in the process table.

Step to reproduce : 

1/ start the cosmian ksm with an proteccio hsm using the --hsm-{model,password,slot} and --hsm admin args.
2/ check the process table with ps auxww | grep hsm-password.

Expected : 

The password should be not in the process table.

Fix : 

Using an environment variable like : HSM_PASSWORD with CLIP.

Best Regards,

### LINKED ISSUE #723: Improve devs documentation
> I kinda agree that we're not lacking folders to clutter the repo and it might be useful to delete this and add a "developers README" with good configs
> 
> I red the README and for now the only thing that should (probably) stay is that CSpell table (if all the team uses vscode... Otherwise we should simply agree on a shared spelling checker
> ( there is another file I don't like called "ltex.dictionary.en-US.txt" )
> 
> I will put an issue with all of these comments 

 _Originally posted by @HatemMn in [#601](https://github.com/Cosmian/kms/pull/601/changes#r2827459564)_

### LINKED ISSUE #739: UI should not block key creation with dev or debug builds
Currently, due to this line in Sidebar.tsx

```tsx
// line 58
                // Handle disabled state based on access rights
                if (isCreateItem || isImportItem) {
                    newItem.disabled = !hasCreateAccess;
                }
```

The UI "grays" out and disables key creation if there is no auth configured, this friction is however unnecessary for developers / debuggers / etc and having to comment out this line on each single debug is problematic

**Side note:** : it might be also interesting to check the behavior of those in a non-debug setting and if the backend successfully refuses non authetificated users

### LINKED ISSUE #763: cosmian kms (rsa|sym) keys destroy wrong type
With the fallowing command you can destroy an AES key : 

`$ cosmian kms rsa keys destroy --key-id hsm::1::master_kek`

I was expecting an error message on this command, because _the master_kek_ is an AES key (You can have an AES, RSA key with the same label in the HSM). The more appropriate command would be : 

`$  cosmian kms sym keys destroy --key-id hsm::1::master_kek`

The type of key is not handled correctly and operation can miss the target.

### LINKED ISSUE #776: Using ckms bench as an regression testing tools
The new kms cli aka ckms as a bench option : 

```
$  cargo r --bin ckms bench -v -b 1000 -n 1000
...
Encryption time XXX XX XXµs => XX 301µs/batch => XXµs/value
Decryption time XXX XX XXµs => XX 648µs/batch => XXµs/value
Amortized encryption time (1 threads): XX XX XXµs => XXX XXXµs/batch => XXXµs/value
Amortized decryption time (1 threads): XX XX XXµs => XXX XXXµs/batch => XXXµs/value
```

It's a perfect tool for **regression testing** : it is a process that ensures that software still performs as expected, even after a **change**.

For that we need a stable and dedicated platform for benchmarking. The hardware must stay the same between versions of the KMS.

### LINKED ISSUE #811: Wildcard client certificate
ACHTUNG, BICYCLETTE

This code in [crate/interfaces/src/hsm/hsm_store.rs](https://github.com/Cosmian/kms/pull/801/changes#diff-7c8788f9ef6aeca1911fe3f4fe88f4b2bdf27b6796c84725a194601203c9bf37) is really dangerous. 

image

The client username is provided by the common name (CN) field from the X.509 certificate.
You can use a wildcard client certificate : 
```
$ cat openssl.cnf
...
[ dn ]
CN = *

[ req_ext ]
subjectAltName = @alt_names

[ alt_names ]
DNS.1 = *
...
$ openssl req -config openssl.cnf -new -x509 -newkey rsa:2048 -nodes -keyout client.key -days 365 -out client.crt
$ openssl x509 -in client.crt -text -noout
...
        Subject: CN=*
...
```

I knew that is very bad and you probably shouldn't use a wildcard client certificate, but there is almost no limitation on the common name ( cf : https://stackoverflow.com/questions/5136198/what-strings-are-allowed-in-the-common-name-attribute-in-an-x-509-certificate ).

_Originally posted by @serene-kitfisto-8899 in https://github.com/Cosmian/kms/issues/801#issuecomment-4090254000_

Manuthor:
# Changelog

All notable changes to this project will be documented in this file.

#### OpenSSH PKCS#11 Support

- **Reliable key material refresh**: fixed `ObjectsStore::upsert()` replacement logic so placeholder objects are properly updated with fetched key bytes, preventing `CKR_GENERAL_ERROR` during OpenSSH key enumeration.
- **Correct public-key decoding paths**: fixed RSA/EC public key extraction to use SPKI BIT STRING payload bytes and refactored conversion through `try_from_spki`, including correct EC OID handling.
- **PKCS#11-compliant EC point export**: encoded `CKA_EC_POINT` as DER OCTET STRING (PKCS#11 v2.40), enabling OpenSSH/OpenSSL parsing compatibility.
- **Safer attribute exposure for mixed key types**: guarded RSA-only attributes (`CKA_MODULUS`, `CKA_PUBLIC_EXPONENT`) behind `is_rsa()` checks to avoid non-RSA lookup failures.
- **Provider runtime and API hardening**: migrated provider internals to lock-free/shared primitives (`OnceLock`, shared runtime, `LazyKeyMaterial`) and reduced cloning/boilerplate (`remote_id() -> &str`, macro-based trait impls), improving stability and performance under OpenSSH PKCS#11 usage patterns.

#### OpenSSH and LUKS Integration

- **test_openssh.sh**: New shell integration test for the OpenSSH PKCS#11 provider; invokes `ssh-keygen -D` to enumerate RSA and EC P-256 public keys via the PKCS#11 library.
- **test_luks.sh**: New shell integration test for the LUKS disk-encryption PKCS#11 scenario (Rust unit tests + shell verification with `ckms locate`).
- **test_luks.sh Part 3**: Added cryptsetup LUKS2 PKCS#11 enrollment integration test — creates a 16 MiB LUKS2 container, enrolls the Cosmian KMS PKCS#11 token via `systemd-cryptenroll`, and verifies enrollment via `cryptsetup luksDump`.

#### Documentation and Integrations

- **Docs**: Reintegrate PKCS#11 pages from `cli_documentation/docs/pkcs11` into main docs under `documentation/docs/integrations`, grouping database integrations in `integrations/databases`, disk encryption in `integrations/disk_encryption`, and adding an OpenSSH integration entry.

#### Server Security and Configuration

- **TLS auth** (#811): Reject client certificates whose CN is empty or `*`; prevents wildcard spoofing attacks.
- **HSM config** (#695): Expose `KMS_HSM_PASSWORD` and `KMS_HSM_SLOT` environment variables for `--hsm-password` / `--hsm-slot` server options so HSM credentials can be injected without config-file edits.

#### CLI Operations

- **CLI destroy type-safety** (#763): `ckms {sym,rsa,ec,pqc,cc} keys destroy` now performs a `GetAttributes` pre-flight check and rejects attempts to destroy a key of the wrong type with a clear error message.

#### HSM Operations

- **Server-side HSM destroy type guard** (#763): When `Destroy.expected_object_type` is set and the target UID belongs to an HSM object (prefix `hsm::`), the server performs a PKCS#11 attribute roundtrip to retrieve the actual key type and rejects the destroy with `Invalid_Object_Type` if the types do not match (e.g. attempting to destroy an AES key via `rsa keys destroy`). ([#763](https://github.com/Cosmian/kms/issues/763))
- **HSM destroy type-guard test assertion** (#763): Fixed `send_message` test helper in HSM tests to include `result_reason` in the error string so that `Invalid_Object_Type` is surfaced when the destroy-type guard fires; the assertion now reliably matches the KMIP `ErrorReason`. ([#763](https://github.com/Cosmian/kms/issues/763))

#### Web UI

- **UI no-auth mode** (#739): The web UI `create` / `import` buttons are now enabled immediately in no-auth mode (`AuthMethod::None`); previously the async sequencing called the permissions API before the auth method was resolved, causing buttons to stay disabled.

#### SBOM Generation

- **Fix sbomnix version and arguments**: The global nixpkgs pin ships an older sbomnix that does not support `--impure` or `--include-vulns`, causing *"unrecognized arguments"* errors in CI. Pinned sbomnix to **v1.7.4** via its own GitHub flake (`github:tiiuae/sbomnix/v1.7.4`) — independent of the nixpkgs pin — so the supported flags are guaranteed. Restored `--impure --include-vulns` on all three `sbomnix` invocations, moved `NIX_CONFIG=nix-command flakes` export to script start (needed for `nix run`), and removed the now-unnecessary `dedup_cves.py` post-processing step.

#### AI Agent Instructions

- **AI agent symlinks** (#723): `AGENTS.md` and `CLAUDE.md` are now symlinks to `.github/copilot-instructions.md` so all AI agents always read the single authoritative instructions file.

#### KMIP Wrapping Documentation

- **`CKM_RSA_AES_KEY_WRAP` invocation** (#688): Document that this scheme is selected by pairing `CryptographicAlgorithm::RSA` with `PaddingMethod::None`; explains the counter-intuitive routing (None ≠ unpadded RSA), adds a KMIP JSON TTLV example, and adds a routing table. Fix broken `../algorithms.md` links in `_export.md` and `_import.md`.

#### Benchmarking and CI Documentation

- **Benchmarks CI** (#776): `benchmarks.sh` now builds the KMS server + ckms CLI, starts a temporary SQLite KMS instance, and runs `ckms bench --speed sanity --format json` as an end-to-end smoke test; supports `BENCH_SAVE_BASELINE` / `BENCH_LOAD_BASELINE` env vars for criterion regression comparisons on a dedicated machine.
- **Benchmark regression workflow** (#776): New `benchmark_regression.sh` script and `benchmark.yml` GitHub Actions workflow provide automated performance regression detection. The script downloads the reference `benchmarks.json` from `package.cosmian.com`, runs benchmarks on the current branch, and fails if the average global regression exceeds a configurable threshold (default 10%). The workflow runs on a self-hosted runner (for stable timings) on a weekly schedule and on demand. ([#776](https://github.com/Cosmian/kms/issues/776))

#### Script Infrastructure Reorganization

- **Script reorganization**: Reorganized 76 scripts from the flat `.github/scripts/`, `nix/scripts/`, and `scripts/` directories into logical subdirectories under `.github/scripts/`: `test/`, `build/`, `package/`, `release/`, `benchmarks/`, `pykmip/`, `sbom/`, `docs/`, `demo/`, `windows/`, `shared/`. All cross-references in `nix.sh`, workflow YAMLs, and the scripts themselves have been updated. Added `shared/colors.sh` for shared terminal color helpers and `benchmarks/docker_helpers.sh` for shared Docker benchmark utilities.
- **Windows workflow script paths**: Fixed `test_windows.yml` and `build_windows.yml` to reference reorganized PowerShell scripts under `.github/scripts/windows/` (`cargo_test.ps1`, `test_ui.ps1`, `cargo_build.ps1`, `windows_ui.ps1`) and updated job display names to use `${{ inputs.archive-name }}`.
- **PyKMIP / Synology DSM REPO_ROOT bug** (#835): Fixed `test_pykmip.sh`, `test_synology_dsm.sh`, and `azure_ekm_mtls_test.sh` to source `common.sh` first then use `get_repo_root` instead of the incorrect `cd "$SCRIPT_DIR/../.."` (resolved to `.github/` instead of repo root), which caused `COSMIAN_KMS_CONF` to point to a non-existent doubled path (`.github/.github/scripts/pykmip/kms.toml`). ([#835](https://github.com/Cosmian/kms/pull/835))
- **REPO_ROOT bug in remaining test scripts** (#835): Fixed same `cd "$SCRIPT_DIR/../.."` REPO_ROOT bug in `test_xks.sh`, `test_gcp_cmek.sh`, and `test_lb_kms_shutdown.sh`; all now source `common.sh` first and use `get_repo_root`. ([#835](https://github.com/Cosmian/kms/pull/835))
- **Stale `build_ui.sh` paths after reorganization**: Fixed `release.sh` and `build_ui_all.sh` which still referenced `build_ui.sh` at the old root path; updated to `build/build_ui.sh` after the script reorganization.
- **`run_benchmarks.sh` local development**: Script now sources `common.sh`, uses `get_repo_root`, and auto-detects `OPENSSL_MODULES`/`OPENSSL_CONF` from the Cargo target directory (with `arm64`→`aarch64` normalization) so it can start the KMS server outside of Nix without a missing-provider error.

Closes #811
Closes #695
Closes #763
Closes #739
Closes #776
Closes #723
Closes #688


cli_documentation/docs/pkcs11/cryhod.md
serene-kitfisto-8899, line 1: Image

_pk = private key ou public key? I'm confused.

Manuthor: Correct. I'll fix it on a next commit.

cli_documentation/docs/pkcs11/veracrypt.md
serene-kitfisto-8899, line 1: TrueCrypt was evaluated by the BSI, alles ist klar ;-)
https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/Studies/Veracrypt/Veracrypt.pdf?__blob=publicationFile&v=1

cli_documentation/docs/pkcs11/openssh.md
serene-kitfisto-8899, line 1: Leaking username bgrieder in the path.

PR #901, opened 2026-04-04
chore: centralize client crates on crate/clients

Manuthor:
maintenance operation


HatemMn:
I think this PR has introduced a small problem (at least for me) and the fact it didn't break the CI is pretty much a mystery

```typescript
const srcPkg = path.join(repoRoot, "crate", "wasm", "pkg");
``` 
Should become

```typescript
const srcPkg = path.join(repoRoot, "crate", "clients", "wasm", "pkg");
``` 

File is `/home/hatem/Desktop/repos/kms/target/debug-fix/ui/scripts/sync-wasm.mjs`

**might** be interesting to add a test ? Idk, maybe not. The error isn't a big issue but it's weird the CI didn't react since it does build wasm

PR #918, opened 2026-04-17
add_oracle_smoketest

chokoblitz:
Adding an automated KMS test to an external Oracle TDE server

### DIFF
--- a/.github/scripts/oracle/smoke-test-tde.sh
+++ b/.github/scripts/oracle/smoke-test-tde.sh
@@ -1,0 +1,396 @@
+#!/usr/bin/env bash
+# smoke-test-tde.sh — TDE proof script for the Oracle KMS integration
+#
+# Usage: bash smoke-test-tde.sh <DEMO_PASS> <TAG_ONLY> <ORACLE_WALLET_PASS> <COSMIAN_HSM_PIN>
+#
+# Runs entirely on the remote Oracle server (copied there via scp).
+# Exits non-zero if any proof fails.
+
+set -euo pipefail
+
+DEMO_PASS="${1:?DEMO_PASS (arg 1) is required}"
+TAG_ONLY="${2:?TAG_ONLY (arg 2) is required}"
+ORACLE_WALLET_PASS="${3:?ORACLE_WALLET_PASS (arg 3) is required}"
+COSMIAN_HSM_PIN="${4:?COSMIAN_HSM_PIN (arg 4) is required}"
+
+ORACLE_HOME=/opt/oracle/product/23ai/dbhomeFree
+ORACLE_SID=FREE
+export ORACLE_HOME ORACLE_SID
+
+PASSED=0
+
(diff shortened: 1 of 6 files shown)


.github/workflows/packaging-docker.yml
```diff
@@ -249,3 +249,72 @@ jobs:
             images+="${tag} "
           done
           cosign sign --yes ${images}
+
+  upgrade-kms-oracle:
```
Manuthor: If you don't mind, remove this Github job in favor of using the Bash script `test_docker_image.sh`. Indeed this script already consumes the `DOCKER_IMAGE_NAME` which is what we need: `ghcr.io/cosmian/kms:xxx` to run the file `.github/scripts/docker-compose.yml`. I believe the KMS container that you're looking for is `no-authentication`.

You could then directly call at the end of test_docker_image.sh your 5 steps:
- Copy upgrade script to oracle
- Run KMS upgrade on oracle
- Copy smoke test script to oracle
- Run TDE smoke test on oracle
- Cleanup smoke test script on oracle

.github/scripts/oracle/smoke-test-tde.sh
```diff
@@ -0,0 +1,396 @@
...
+# Usage: bash smoke-test-tde.sh <DEMO_PASS> <TAG_ONLY> <ORACLE_WALLET_PASS> <COSMIAN_HSM_PIN>
+#
+# Runs entirely on the remote Oracle server (copied there via scp).
+# Exits non-zero if any proof fails.
+
+set -euo pipefail
+
+DEMO_PASS="${1:?DEMO_PASS (arg 1) is required}"
+TAG_ONLY="${2:?TAG_ONLY (arg 2) is required}"
+ORACLE_WALLET_PASS="${3:?ORACLE_WALLET_PASS (arg 3) is required}"
```
Manuthor: Why do we need the secret ORACLE_WALLET_PASS? It could be hardcoded for test purposes, no?

.github/scripts/oracle/smoke-test-tde.sh
```diff
@@ -0,0 +1,396 @@
...
+COSMIAN_HSM_PIN="${4:?COSMIAN_HSM_PIN (arg 4) is required}"
```
Manuthor: COSMIAN_HSM_PIN is misleading since it is the bearer token when `pkcs11_use_pin_as_access_token = true` is enabled: see https://docs.cosmian.com/key_management_system/integrations/databases/oracle_tde/#hsm_identity_pass-and-authentication-modes.

You can just hardcode the `XXX` value since KMS server will not be started with real OIDC authentication (it should first require a login on the OIDC to retrieve a fresh access token which is not easily possible in CI)

Manuthor (commented):
Great, thanks.
However could you please verify that new Oracle tests are executed? I do not find them in https://github.com/Cosmian/kms/actions/runs/25165777938/job/73771431629?pr=918

PR #932, opened 2026-04-28
feat(config): add secret management for KMS config files

### LINKED ISSUE #882: Configuration secret management: no clear-text secrets in TOML
## Summary

The current configuration file contains plain-text secrets (HSM slot passwords, database credentials, certificate passwords, LDAP bind passwords). Sharing this file across teams violates least-privilege principles and significantly increases the blast radius of a credential compromise.

This was raised as a field blocker preventing enterprise onboarding in regulated environments.

## Impact

- Security teams in regulated industries reject configurations that store secrets in plain text
- Single config file must be shared across ops, DBA, and integration teams, exposing all secrets to all teams
- No mechanism to inject secrets from external secret stores

## Implementation plan

1. **Split config** into two files (backward-compatible — single file still works):
   - `server.toml` — network, TLS, server identity, feature flags (shareable)
   - `secrets.toml` — HSM passwords, DB credentials, integration tokens (root-only)
2. **Environment variable interpolation**: support `password = "${KMS_HSM_PASSWORD}"` syntax in all TOML string values, resolved at startup
3. **Phase 2**: native `vault://`, `aws-ssm://`, `azure-kv://` URI scheme resolution for any TOML string value

**Files**: `crate/server/src/config/`, `resources/server.toml`, `pkg/kms.toml`
**Effort**: Small (1-2 weeks) | No prerequisites

### LINKED ISSUE #951: Use of clap_config_fallback
The goal of this task is to consolidate the way the TOML configuration or clap arguments are given to KMS server of KMS CLI using this crate https://crates.io/crates/clap_config_fallback.

p0wline:
Closes #882
Closes #951 

## What

Add secret management support to KMS TOML configuration files so that sensitive values (passwords, tokens, API keys) are never stored in clear text.

### Phase 1 — Local secret handling
- **Env var interpolation**: `${VAR_NAME}` and `${VAR_NAME:-default}` syntax resolved at startup in all config values
- **Secrets file**: split sensitive values into a separate TOML file, referenced via `secrets_file` key in the main config or `COSMIAN_KMS_SECRETS_CONF` env var; deep-merged at startup before deserialization
- New `pkg/secrets.toml` template with documented examples

### Phase 2 — External secret backends (optional feature flags)
URI schemes resolved at startup:

| URI scheme | Feature flag | Backend |
|---|---|---|
| `vault://<mount>/<path>[#<field>]` | `secret-vault` | HashiCorp Vault KV-v2 |
| `aws-ssm://<region>/<param-path>` | `secret-aws` | AWS SSM Parameter Store (SecureString) |
| `azure-kv://<vault>/secrets/<name>` | `secret-azure` | Azure Key Vault REST API |

Each backend spawns a dedicated OS thread with its own Tokio runtime to avoid nested-runtime panics at startup.

## Testing
- Unit tests: env var interpolation, secrets file deep-merge, loading via config key and env var
- End-to-end tested: Vault ✅, AWS SSM ✅, Azure KV ✅


crate/server/Cargo.toml
Manuthor, line 1: Not sure about the changes made on this file since the reordering blocks the comparison. Unless really needed, revert them please.

.github/workflows/test_all.yml
```diff
@@ -134,8 +155,8 @@ jobs:
     # give them fixed concurrency groups so only one job runs at a time across all PRs.
     # utimaco and softhsm2 use a per-run group so they are never blocked by other PRs.
     concurrency:
-      group: ${{ (matrix.hsm-type == 'proteccio' && 'hsm-proteccio') || (matrix.hsm-type == 'crypt2pay' && 'hsm-crypt2pay') || format('hsm-{0}-{1}', matrix.hsm-type,
-        github.run_id) }}
+      group: ${{ matrix.hsm-type == 'proteccio' && 'hsm-proteccio' ||
```
Manuthor: Revert please

.github/scripts/nix.sh
```diff
@@ -509,6 +509,29 @@ test_command() {
...
+    ;;
+  secret_aws)
+    SCRIPT="$REPO_ROOT/.github/scripts/test/test_secret_aws.sh"
+    for var in AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_REGION; do
+      if [ -z "${!var:-}" ]; then
+        echo "Error: Required environment variable $var is not set" >&2
+        echo "AWS SSM secret backend tests require AWS credentials." >&2
+        exit 1
+      fi
+    done
```
Manuthor: Shouldn't it be in this order?

```suggestion
    for var in AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_REGION; do
      if [ -z "${!var:-}" ]; then
        echo "Error: Required environment variable $var is not set" >&2
        echo "AWS SSM secret backend tests require AWS credentials." >&2
        exit 1
      fi
    done
    SCRIPT="$REPO_ROOT/.github/scripts/test/test_secret_aws.sh"
```

.github/scripts/nix.sh
```diff
@@ -509,6 +509,29 @@ test_command() {
...
+    ;;
+  secret_azure)
+    SCRIPT="$REPO_ROOT/.github/scripts/test/test_secret_azure.sh"
+    for var in AZURE_TENANT_ID AZURE_CLIENT_ID AZURE_CLIENT_SECRET AZURE_KV_NAME; do
+      if [ -z "${!var:-}" ]; then
+        echo "Error: Required environment variable $var is not set" >&2
+        echo "Azure KV secret backend tests require Azure credentials." >&2
+        exit 1
+      fi
+    done
```
Manuthor: Same here

deny.toml
```diff
@@ -72,7 +70,10 @@ ignore = [
...
   { id = "RUSTSEC-2025-0010", reason = "Pinned to scratchstack-aws-signature 0.10 (Feb 2026) which depends on ring 0.16.x" },
   { id = "RUSTSEC-2023-0071", reason = "Marvin Attack: potential key recovery through timing sidechannels. KMS depends now on jsonwebtoken that has 2 crypto backends (rust_crypto and aws-lc). aws-lc not being Windows compatible, we must choose rust_crypto" },
-] # The path where the advisory databases are cloned/fetched into
+  { id = "RUSTSEC-2026-0098", reason = "rustls-webpki 0.101.7 via hyper-rustls 0.24 / rustls 0.21 (hyper 0.14 dependency). URI name constraints bypass requires certificate misissuance; KMS does not assert URI SANs." },
+  { id = "RUSTSEC-2026-0099", reason = "rustls-webpki 0.101.7 via hyper-rustls 0.24 / rustls 0.21 (hyper 0.14 dependency). Wildcard name constraint bypass requires misissuance to exploit." },
+  { id = "RUSTSEC-2026-0104", reason = "rustls-webpki 0.101.7 via hyper-rustls 0.24 / rustls 0.21 (hyper 0.14 dependency). CRL parsing panic is not reachable: KMS does not use certificate revocation lists." },
+]  # The path where the advisory databases are cloned/fetched into
```
Manuthor: Can we avoid this whitelisting?

.cargo/audit.toml
```diff
@@ -9,4 +9,10 @@ ignore = [
   "RUSTSEC-2025-0009",
   "RUSTSEC-2025-0010",
   "RUSTSEC-2023-0071",
+  # rustls-webpki 0.101.7 via hyper-rustls 0.24 / rustls 0.21 (hyper 0.14 transitive dependency).
+  # KMS does not use CRLs, does not assert URI SANs, and exploiting wildcard name constraints
+  # requires certificate misissuance upstream.
+  "RUSTSEC-2026-0098",
+  "RUSTSEC-2026-0099",
+  "RUSTSEC-2026-0104",
```
Manuthor: Can we avoid this?

Manuthor (changes requested):
Thanks for this!

Only a partial review.
Could you rebase please?

deny.toml
p0wline: It resolved CI issues but we can try again yes.

.cargo/audit.toml
p0wline: Same than the previous file.

Manuthor:
We should also support our own KMS server for storing this kind of config files secrets and support another `URI scheme` for a Cosmian KMS

pkg/secrets.toml
Manuthor, line 1: Do we really need this new file? 
Could we just reuse the same old kms.toml instead? (to avoid packaging issues)

crate/server/Cargo.toml
Manuthor, line 1: Please revert changes on this file

CHANGELOG/.gitignore
Manuthor, line 1: Can we keep this .gitignore file? (git refuses to commit empty folder)

.github/workflows/test_all.yml
```diff
@@ -33,6 +33,9 @@ jobs:
           - gcp_cmek
           - openssh
           - luks
+          - secret_vault
+          - secret_aws
+          - secret_azure
```
Manuthor, line 41: Can we merge the 3 into 1 `conf_secrets` enum and add an additional bash script to call test_secret_xxx.sh?

crate/server/src/config/command_line/clap_config.rs
```diff
@@ -215,6 +215,81 @@ pub struct ClapConfig {
     pub kmip_policy: KmipPolicyConfig,
 }
 
+/// Interpolates `${VAR_NAME}` placeholders in a string with environment variable values.
+///
+/// For each occurrence of `${VAR_NAME}`, the value of the environment variable `VAR_NAME`
+/// is substituted. Returns an error message if a referenced variable is not set.
+fn interpolate_env_vars(content: &str) -> Result<String, String> {
```
Manuthor: I do not understand the logic of the 2 new functions and their implementation seem to be fragile to future changes.
Moreover the existing function `default_config_with_comments` is unnecessary complicated. Could you resorb this complexity by solving this issue https://github.com/Cosmian/kms/issues/951?

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.91.0"
+channel = "1.94.1"
```
Manuthor: Please keep this change for later in a separate PR

deny.toml
```diff
@@ -103,7 +101,8 @@ allow = [
   "Unicode-3.0",
   "CC0-1.0",
   "BUSL-1.1",
-  "CDLA-Permissive-2.0"
+  "GPL-3.0-or-later",  # legacy dependency note
```
Manuthor: We want to get rid of this license GPL-3.0-or-later

crate/server/Cargo.toml
```diff
@@ -48,11 +48,21 @@ non-fips = [
...
 # Enable interop test operations (should only be used for testing)
 interop = ["cosmian_kms_server_database/interop"]
+# Secret backend: HashiCorp Vault KV-v2
+secret-vault = []
+# Secret backend: AWS SSM Parameter Store
+secret-aws = ["dep:hmac", "dep:sha2"]
+# Secret backend: Azure Key Vault
+secret-azure = []
+# Secret backend: Cosmian KMS
+secret-cosmian-kms = []
```
Manuthor: Do not add new features please, unless really necessary

crate/server/src/config/command_line/clap_config.rs
```diff
@@ -210,11 +210,122 @@ pub struct ClapConfig {
...
 }
 
+/// Interpolates `${VAR}` and `${VAR:-default}` placeholders in a single string value.
+///
+/// Called on each string leaf of the TOML tree after the document has been parsed,
+/// so TOML comments have already been stripped and TOML-injection is not possible.
+///
+/// `${VAR}` — substitutes the value of `VAR`; returns an error if `VAR` is not set.
+/// `${VAR:-default}` — substitutes the value of `VAR` if set, otherwise uses `default`.
+pub(super) fn interpolate_string_value(s: &str) -> Result<String, String> {
```
Manuthor: Can we get rid of this new functions?

rust-toolchain.toml
p0wline: This bump is required by `clap-config-fallback` v0.1.11 whose MSRV is 1.94.1 (see https://crates.io/crates/clap-config-fallback). Without it the crate won't compile. Since issue #951 explicitly asks us to adopt this crate, the toolchain bump is a direct prerequisite — I don't think we can defer it to a separate PR. Happy to discuss if you have a different approach in mind.

deny.toml
p0wline: Done — removed in this commit. No dependency in the workspace actually carries a GPL-3.0-or-later license; this line was added defensively and is not needed.

crate/server/Cargo.toml
p0wline: Agreed — the `secret-cosmian-kms` feature (using a Cosmian KMS as a secret store via `cosmian-kms://` URIs) was not part of issue #882's original scope. Removed in this commit.

crate/server/src/config/command_line/clap_config.rs
p0wline: These four functions are the core feature of this PR and directly address issue #882. Here is what each one does and why it cannot be removed:

- `interpolate_string_value` / `interpolate_toml_value` / `parse_toml_and_interpolate`: implement the `${ENV_VAR}` and `${ENV_VAR:-default}` substitution syntax in TOML string values. Issue #882 explicitly requests this so that secrets like HSM passwords and DB credentials can be injected at startup from the environment instead of stored in plain text.

- `deep_merge_toml`: merges a separate `secrets.toml` file (referenced via `secrets_file` in the main config or `COSMIAN_KMS_SECRETS_CONF` env var) into the main config before deserialization. This is point 1 of the implementation plan in #882 (split config into `server.toml` + `secrets.toml`).

`clap-config-fallback` (issue #951) handles a different concern: merging CLI arguments with a TOML config file. It does not provide env-var interpolation inside TOML values or a secrets-file merge mechanism.

If we remove these functions, the PR no longer resolves #882. Happy to discuss an alternative approach if you had something else in mind.

crate/server/src/config/command_line/clap_config.rs
p0wline: WIP

CHANGELOG/.gitignore
Manuthor, line 1: Do not delete this file please

.github/workflows/test_all.yml
```diff
@@ -150,18 +162,30 @@ jobs:
...
           set -ex
           bash .github/scripts/nix.sh --variant ${{ matrix.features }} test ${{ matrix.type }}
 
   hsm:
     name: HSM ${{ matrix.hsm-type }} - ${{ matrix.features }}
     runs-on: ubuntu-latest
-    # proteccio and crypt2pay hardware do not support concurrent connections from multiple CI runs;
-    # give them fixed concurrency groups so only one job runs at a time across all PRs.
+    # proteccio hardware does not support concurrent connections from multiple CI runs;
+    # give it a fixed concurrency group so only one job runs at a time across all PRs.
```
Manuthor: Is the change in comment done on purpose? Because the concurrency rule also applies on crypt2pay.

.github/reusable_scripts
Manuthor, line 1: Is the change mandatory? If not, revert it (`git checkout origin/develop .github/reusable_scripts`)

crate/server/src/config/command_line/clap_config.rs
```diff
@@ -155,37 +155,37 @@ pub struct ClapConfig {
     #[clap(verbatim_doc_comment, long, env = "KMS_PUBLIC_URL")]
     pub kms_public_url: Option<String>,
 
-    #[clap(flatten)]
+    #[command(flatten)]
```
Manuthor, line 160: For my knowledge, what is the diff between #[clap(flatten)] and #[command(flatten)]

crate/server/src/config/command_line/clap_config.rs
Manuthor: Sorry to insist but I suggest we remove this feature (and those interpolate_xxx, parse_toml_and_interpolate, and deep_merge_toml functions) that could involve a security issue. And this is not Windows compliant. I suggest we do not support interpolating env. variable in the URL given in kms.toml. URLs must be written as full evaluated string without environment variables in it.

crate/server/src/config/command_line/clap_config.rs
```diff
@@ -716,10 +878,12 @@ mod tests {
 
         // Save current env state
         let original_env = std::env::var("COSMIAN_KMS_CONF").ok();
+        let original_secrets_env = std::env::var("COSMIAN_KMS_SECRETS_CONF").ok();
 
         // Clear env
         unsafe {
             std::env::remove_var("COSMIAN_KMS_CONF");
+            std::env::remove_var("COSMIAN_KMS_SECRETS_CONF");
```
Manuthor:
```suggestion
```

crate/server/src/config/command_line/secret_backends.rs
```diff
@@ -0,0 +1,1012 @@
...
+//! Supported schemes and their controlling feature flags:
+//!
+//! | Scheme             | Feature flag          | Required env vars / notes                  |
+//! |--------------------|----------------------|--------------------------------------------|
+//! | `vault://`         | `secret-vault`       | `VAULT_ADDR`, `VAULT_TOKEN`                |
+//! | `aws-ssm://`       | `secret-aws`         | `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` |
+//! | `azure-kv://`      | `secret-azure`       | `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`,      |
+//! |                    |                      | `AZURE_CLIENT_SECRET`                      |
+//! | `cosmian-kms://`   | _(always enabled)_   | `COSMIAN_KMS_TOKEN` (opt.),                |
+//! |                    |                      | `COSMIAN_KMS_INSECURE_CERTS` (opt.)        |
```
Manuthor: IMHO we need to avoid the interpolation of environment variables in Rust source code and the parsing of them being a security issue and hard to maintain. 
I suggest we use clap in command line argument (or external environment variables) to retrieve the full URI of those secret backends as user inputs.
Please tell copilot to greatly simplify this current implementation by using Rust design patterns such as Generics, traits or shared functions.

crate/server/src/config/command_line/ui_config.rs
```diff
@@ -55,7 +63,8 @@ pub struct UiConfig {
     #[clap(verbatim_doc_comment, long, env = "KMS_SESSION_SALT")]
     pub ui_session_salt: Option<String>,
 
-    #[clap(flatten)]
+    #[config(no_flatten)]
+    #[command(flatten)]
```
Manuthor, line 74: I do not get this. Isn't it contradictory?

crate/server/src/tests/rest_crypto/jose_vectors.rs
```diff
@@ -183,8 +183,15 @@ where
...
-    let (kid, _) = provision_key(app, key).await?;
+    // RFC JWKs may omit the optional `alg` field. Fill it in from the vector's
+    // top-level `algorithm` so the key-import endpoint knows which algorithm to use.
+    let mut key = v["key"].clone();
+    if key.get("alg").is_none_or(Value::is_null) {
+        if let Some(alg) = v["algorithm"].as_str() {
+            key["alg"] = Value::String(alg.to_owned());
+        }
+    }
+    let (kid, _) = provision_key(app, &key).await?;
```
Manuthor: Why this change?

documentation/docs/favicon.png
Manuthor, line 1: Revert.

default.nix
Manuthor, line 1: Revert or explain

Manuthor (commented):
In addition to these new comments, could you please squash all your commits in 1 commit (to make sure the rebase with origin/develop is correct)

crate/server/src/config/command_line/ui_config.rs
p0wline: `#[config(no_flatten)]` tells `clap_config_fallback` NOT to flatten this nested struct into the parent TOML table: the KMS config file must use a `[ui_config.ui_oidc_auth]` sub-table.

`#[command(flatten)]` is a clap directive that inlines CLI flags from`OidcConfig` into the parent struct; it is unrelated to TOML layout.

The two attributes operate at different layers (config file vs CLI) and are intentionally used together.

crate/server/src/tests/rest_crypto/jose_vectors.rs
p0wline: According to Claude :
RFC JWKs are spec-compliant in omitting alg, but the test harness needed it to call the import API. Without this fix, JOSE tests fail on "pure" RFC vectors.

Should I revert ?

Manuthor (approved):

Issue #881, opened 2026-04-01
Structured audit trail and SIEM integration


Manuthor:
## Summary

Every key operation should emit a structured JSON audit event so that security teams can feed events into their SIEM, detect anomalies, and produce compliance evidence.

## Impact

- Without structured audit events, regulated customers cannot satisfy PCI-DSS, HIPAA, or SOC 2 audit requirements
- Absence of SIEM integration is a blocker for security-conscious enterprise accounts
- OpenTelemetry traces exist but do not provide a per-operation, user-attributed audit record

## Implementation plan

1. New middleware `crate/server/src/middlewares/audit.rs` emitting structured JSON per request:
   `{ timestamp, operation, user, object_uid, algorithm, client_ip, result, duration_ms }`
2. Export targets:
   - OpenTelemetry log export (already wired infrastructure)
   - Syslog CEF format (ArcSight / QRadar)
   - HTTP Event Collector endpoint (Splunk HEC)
3. New config block `[audit]` with per-target enable flags and endpoint configuration
4. CLI: `ckms audit export --since <date> --format json|cef`

**Effort**: Medium (3-4 weeks) | No prerequisites

HatemMn:
## What exists on the KMS :

1. Aggregate OTel metrics per (user, operation)	
2. Unstructured diagnostic logging to syslog/file	
3. Authenticated user identity per request	
4. OTLP traces export

**What does not exist :**

1. Per-operation structured event with user + outcome + object_uid	
2. Client IP captured per operation	
3. Operation duration per event (not just aggregate histogram)	
4. Result outcome (success / auth-failure / crypto-error)	
5. Persistent audit event store (DB table or append-only file)	
6. OTel logs signal (distinct from traces/metrics)
7. Syslog CEF export	
8. Splunk HEC export
9. [audit] config section	
10. ckms audit CLI

PR #934, opened 2026-04-29
feat: Structured audit layer middleware

### LINKED ISSUE #1064: Phase 1:  PART B: OTEL should not expose user info
Otel logs should NOT expose usernames

Audit only should

This issue is simply for tracking purposes

### LINKED ISSUE #1067: Phase 1: PART A: base audit on file, offline
This issue covers Phase 1, part A, only: the audit event pipeline and local storage. SIEM export (OTel log signal, Splunk HEC, syslog/CEF) is tracked separately.

# The text below is for reference only, actual code might have diverged to adapt to more realistic solutions. All of this issue was written before the project has started

## What gets built

1. AuditEvent struct

```rust
pub struct AuditEvent {
    pub id:           u64,
    pub timestamp:    OffsetDateTime,
    pub operation:    String,       // e.g. "Encrypt", "Create", "Destroy"
    pub user:         String,       // from AuthenticatedUser or "unauthenticated"
    pub object_uid:   Option<String>,
    pub algorithm:    Option<String>,
    pub client_ip:    Option<String>,
    pub result:       AuditResult,  // Success | Failure(code)
    pub duration_ms:  u64,
    pub prev_hash:    [u8; 32],     // SHA-256 of previous event canonical bytes
    pub row_hash:     [u8; 32],     // SHA-256 of this event (excluding row_hash)
}
```

2. Audit middleware: crate/server/src/middlewares/audit.rs

- Follows the EnsureAuth structural pattern (Transform + Service)
- Registered as the outermost .wrap() on default_scope in start_kms_server.rs
- Reads AuthenticatedUser from request extensions (falls back to "unauthenticated" for 401s)
- Captures response status after service.call(req).await to determine outcome
- Feeds completed AuditEvent into the configured sinks

3. [audit] config section :
```toml
text
[audit]
enabled = true

[audit.file]
enabled = true
path = "/var/log/cosmian_kms/audit.log"  # default if not set

[audit.db]
enabled = false
url = "postgresql://audit_writer:secret@localhost/kms_audit"

[audit.otlp], [audit.splunk_hec], [audit.syslog] slots are reserved but not implemented here.
```

4. Storage backends
out of scope

5. ckms audit CLI subcommands

- `ckms audit export --since <date> --format json|cef # reads from the local audit store`
- `ckms audit verify # walks the hash chain, reports first broken link`

## Out of scope for this issue

- OTel log signal export
- Splunk HEC export
- Syslog/CEF export
- auth_method field on AuthenticatedUser (noted as a possible future extension)

## Acceptance criteria

Schema & payload safety

- [ ] The `AuditEvent` struct captures: who (`user`), what (`operation`), when (`timestamp`), where (`client_ip`), on what (`object_uid`), and with what outcome (`result`, `duration_ms`) — all fields must be populated or explicitly `None` with a documented reason
- [ ] Audit logs must not contain sensitive payloads: no plaintext key material, no credentials, no secrets in any logged field (OWASP A09, ISO 27001 A.8.15)

Runtime behavior

- [ ] All operations passing through `default_scope` produce an `AuditEvent`
- [ ] Auth failures (401) produce an event with `user = "unauthenticated"` and the client IP
- [ ] An audit store write failure logs an `ERROR` in diagnostic logs but does not return an error to the client

Storage & tamper-evidence

- [ ] `[audit.db]` and the key DB are demonstrably separate — different connection strings in test setup
- [ ] The KMS process used for audit storage only appends — UPDATE and DELETE on the audit store are not issued by the KMS Rust source code; if INSERT-only DB privileges are not enforceable by the KMS itself (e.g. file backend), compliance is accepted by code review of the open-source Rust source
- [ ] Hash chain is intact after a sequence of events and verifiable with `ckms audit verify`
- [ ] `ckms audit verify` exits non-zero and identifies the broken link when any past event is modified or deleted

CLI & export

- [ ] `ckms audit export --format json` produces valid, parseable JSON
- [ ] `ckms audit export --format cef` produces valid CEF-formatted lines on stdout
- [ ] `ckms audit export` output must not expose more personal data than what was logged — no enrichment, no joins against user directories at export time (GDPR/RGPD compliance)

Configuration & compliance defaults

- [ ] KMS starts successfully with no `[audit]` section in config — audit is disabled by default, no error
- [ ] When `[audit] enabled = true` with no further config, KMS writes to a stable default path without error
- [ ] Retention period is configurable per storage backend; 6–12 months is documented as the default for EU/FR deployments (CNIL recommendation, NIS2)
- [ ] `[audit.otlp]`, `[audit.splunk_hec]`, `[audit.syslog]` sub-tables are reserved in the config schema — Phase 2 can add them without breaking existing configs

___

#### Other info

Write AuditEvent to a dedicated local store (Postgres/SQLite/MySQL, **separate from key storage**)

**Hash chaining for tamper-evidence**

`ckms audit export --since <date> --format json|cef` reads from this local store

This alone satisfies most of the compliance requirement: you have a durable, tamper-evident, user-attributed record of every key operation

What is marked in bold is not a suggestion but a **requirement** for this issue approval - not satisfying that can fail the Audit phase in the future

**It's also mandatory (for audit compliance) that the KMS is proven to ONLY write logs, and not backward tamper with them**

HatemMn:
## About merging

To keep everything consistent, the following PRs should be rebased on top of each other **in this order**

1. https://github.com/Cosmian/kms/pull/934
2. https://github.com/Cosmian/kms/pull/1063
3. Database PR: coming soon
4. Siem PR: later

---

## Specs (optional) : 

- [x] PCI-DSS compliance :  **important, won't drop it** ( source : check PCI-DSS v4.0.1 Req. 10 )
- [x] HIPPA Compliance :  ( Source : https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-66r2.pdf - section 5.3.2. Audit Controls (§ 164.312(b)) )

___

### Artifacts : 

For testing purposes, these debug artifacts are shared : 

**no_auth_audit.toml**

```toml
default_username = "admin"

[http]
port = 9998
hostname = "0.0.0.0"

[db]
database_type = "sqlite"
sqlite_path = "/tmp/kms-data-e2e"
clear_database = true
unwrapped_cache_max_age = 15

[logging]
rust_log = "info,cosmian_kms=info"
ansi_colors = false

[ui_config]
ui_index_html_folder = "ui/dist"

[audit]
enabled = true

[audit.file]
path = "/tmp/kms-audit-e2e.jsonl"

```

The following commands (all assume you already ran a KMS) : 

```shell
# Terminal 2 (while server is running)
cargo run -p ckms --features non-fips -- --url http://127.0.0.1:9998 sym keys create --algorithm aes --number-of-bits 256 --tag audit-e2e-key
cargo run -p ckms --features non-fips -- --url http://127.0.0.1:9998 sym encrypt -k d5d986ac-ff6c-472c-904e-0b0bf6bfaa92 /tmp/audit-plain.txt
curl -s -o /tmp/kmip-empty.json -w "KMIP_EMPTY_HTTP=%{http_code}\n" -X POST -H "Content-Type: application/json" -d '{}' http://127.0.0.1:9998/kmip/2_1

cargo run -p ckms --features non-fips -- --url http://127.0.0.1:9998 sym keys create --algorithm aes --number-of-bits 256 --tag audit-e2e-key-r2
cargo run -p ckms --features non-fips -- --url http://127.0.0.1:9998 sym encrypt -k 417fe2de-827d-48d0-8d51-851bec315b76 /tmp/audit-plain.txt
cargo run -p ckms --features non-fips -- --url http://127.0.0.1:9998 sym decrypt -k 417fe2de-827d-48d0-8d51-851bec315b76 /tmp/audit-plain.enc
cargo run -p ckms --features non-fips -- --url http://127.0.0.1:9998 sym decrypt -k 00000000-0000-0000-0000-000000000000 /tmp/audit-plain.enc
curl -s -o /tmp/kmip-empty-2.json -w "KMIP_EMPTY_HTTP=%{http_code}\n" -X POST -H "Content-Type: application/json" -d '{}' http://127.0.0.1:9998/kmip/2_1
```

Normally produce **kms-audit.jsonl** :

```json
{"id":0,"timestamp":"2026-07-20T19:02:25.605113314Z","operation":"Create","user":"admin","object_uid":"359019d8-1543-4e2e-9d96-674dd64fcffc","algorithm":"AES","client_ip":"127.0.0.1","result":"Success","duration_ms":4,"request_id":"4b2e6994-0ef9-4dfe-9804-0da4c5edde4b","prev_hash":"0000000000000000000000000000000000000000000000000000000000000000","row_hash":"641c442edc2bb075e9770a905c843fdadfb85503464e67c1658549624d44fb00"}
{"id":1,"timestamp":"2026-07-20T19:02:36.704410987Z","operation":"Encrypt","user":"admin","object_uid":"359019d8-1543-4e2e-9d96-674dd64fcffc","algorithm":"AES","client_ip":"127.0.0.1","result":"Success","duration_ms":12,"request_id":"3618ade8-5db7-4635-9d05-5af6a7614d52","prev_hash":"641c442edc2bb075e9770a905c843fdadfb85503464e67c1658549624d44fb00","row_hash":"637cd569144c90f0ee73564286f8a75c56084cf76f7375b3d31266161b3de6a3"}
{"id":2,"timestamp":"2026-07-20T19:02:43.474676849Z","operation":"kmip","user":"admin","object_uid":null,"algorithm":null,"client_ip":"127.0.0.1","result":{"Failure":"422 Unprocessable Entity"},"duration_ms":0,"request_id":"029b157a-4d66-4aa8-a504-dab433baa374","prev_hash":"637cd569144c90f0ee73564286f8a75c56084cf76f7375b3d31266161b3de6a3","row_hash":"3b00ac7a2880cfe79c30763b3e6e18c37a8019dbe1687d6a1e4936ddd4772695"}
{"id":3,"timestamp":"2026-07-20T19:02:50.496997982Z","operation":"Create","user":"admin","object_uid":"021c29e9-8017-42d6-9887-bebb5c413e74","algorithm":"AES","client_ip":"127.0.0.1","result":"Success","duration_ms":2,"request_id":"2e5fcc19-2021-45a3-822c-9ecf4e38de85","prev_hash":"3b00ac7a2880cfe79c30763b3e6e18c37a8019dbe1687d6a1e4936ddd4772695","row_hash":"b75759b556141c03e2e4c9b6cfb4e8992f3de09338bb32b46eb549b0e1a6f30c"}
{"id":4,"timestamp":"2026-07-20T19:03:00.273344336Z","operation":"Encrypt","user":"admin","object_uid":"021c29e9-8017-42d6-9887-bebb5c413e74","algorithm":"AES","client_ip":"127.0.0.1","result":"Success","duration_ms":2,"request_id":"330d8643-91a4-4869-af8c-849ee50d5643","prev_hash":"b75759b556141c03e2e4c9b6cfb4e8992f3de09338bb32b46eb549b0e1a6f30c","row_hash":"c849c7efd68569ced3b63af96af2ad5d77a82051e36c64c2ef9f5be7eeea17e9"}
{"id":5,"timestamp":"2026-07-20T19:03:00.685874574Z","operation":"Decrypt","user":"admin","object_uid":"021c29e9-8017-42d6-9887-bebb5c413e74","algorithm":"AES","client_ip":"127.0.0.1","result":"Success","duration_ms":0,"request_id":"29df4906-b8c2-47fb-b88b-b8ca3a0c6ceb","prev_hash":"c849c7efd68569ced3b63af96af2ad5d77a82051e36c64c2ef9f5be7eeea17e9","row_hash":"fbca602d562d8205315085ddf9a5493eee14be73180ff79c023061027ea6cde9"}
{"id":6,"timestamp":"2026-07-20T19:03:01.069844169Z","operation":"Decrypt","user":"admin","object_uid":"00000000-0000-0000-0000-000000000000","algorithm":"AES","client_ip":"127.0.0.1","result":{"Failure":"422 Unprocessable Entity"},"duration_ms":1,"request_id":"7a297435-64f3-41ef-a0b2-df64146915de","prev_hash":"fbca602d562d8205315085ddf9a5493eee14be73180ff79c023061027ea6cde9","row_hash":"ad83e2e8730ed5556bf51f2d8b301167cbe54b1314a2a1ce720a274b35f84594"}
{"id":7,"timestamp":"2026-07-20T19:03:01.079329139Z","operation":"kmip","user":"admin","object_uid":null,"algorithm":null,"client_ip":"127.0.0.1","result":{"Failure":"422 Unprocessable Entity"},"duration_ms":0,"request_id":"236a3050-940d-4005-8cbd-f25a4a33627a","prev_hash":"ad83e2e8730ed5556bf51f2d8b301167cbe54b1314a2a1ce720a274b35f84594","row_hash":"a361c07f62b719755b3e5ff687049db728e0c1cdb722a8e3060369f17dd01610"}


```

Then you can run : 

```shell
cargo run -p ckms --features non-fips -- audit verify --path #your file
```

# Other technical considerations

### Reverse proxy usage : 

```rust
    /// Comma-separated list of trusted reverse-proxy IP addresses or CIDR blocks
    /// (e.g. `"10.0.0.0/8,172.16.0.0/12"`).  The `X-Forwarded-For` header is only
    /// trusted when the actual TCP peer address falls within one of these ranges.
    ///
    /// When empty (the default), `client_ip` is always the direct TCP peer address.
    /// Single IPs can be expressed as `/32` (IPv4) or `/128` (IPv6).
    #[clap(
        long = "audit-trusted-proxy-cidrs",
        env = "KMS_AUDIT_TRUSTED_PROXY_CIDRS",
        value_delimiter = ',',
        verbatim_doc_comment
    )]
    #[serde(rename = "trusted_proxy_cidrs", default)]
    pub audit_trusted_proxy_cidrs: Vec<IpNet>,
```

This parameter seems over engibneered, but : 

Threat without allowlist, two bad options:

Trust XFF always → any attacker sends fake X-Forwarded-For: 1.2.3.4 header directly to KMS → audit log records wrong [client_ip]. Forensic/compliance trail poisoned — attacker hides real origin or frames innocent IP. This is audit-log spoofing (CWE-117-ish), works via any protocol (UI, CLI, curl), no UI needed.
Never trust XFF → when real proxy present, [client_ip] always = proxy's own IP → useless for forensics regardless of caller.
CIDR allowlist fixes both: trust XFF only from known proxy IPs, so external attacker can't inject fake header, but real proxy's forwarded value still recorded.

---

# What's audited, what's not

## Audited

- `POST/GET /kmip` — `kmip::kmip_2_1_json` (JSON) and `kmip::kmip` (binary TTLV)
- `access::*` — grant/revoke/list access rights, list owned objects, get current user
- `get_server_info`, `get_hsm_status`, `/download-cli`

## Not audited

- **Raw TCP socket server** — `start_socket_server` → `handle_ttlv_bytes` (start_kms_server.rs). Full KMIP protocol, no `AuditMiddleware` possible (not Actix), no manual audit call either.
- **`/v1/crypto` REST Native Crypto API** (jose/) — `POST /encrypt`, `/decrypt`, `/sign`, `/verify`, `/mac`, `POST /keys` (create), `DELETE /keys/{kid}` (destroy), `POST /keys/unwrap`, tag add/remove/list.
- **Vault-compatible Transit + PKI API** — `/v1/{transit_mount}` (`create_transit_key`, `configure_transit_key`, `delete_transit_key`, `sign_with_transit_key`, ...) and `/v1/{pki_mount}` (`sign_intermediate`), gated by `vault_api_enabled`. `/v1/auth` proxy scope also unaudited (pure forward, no key ops).
- **`/tokenize`** (non-fips only) — `hash`, `noise`, `word_mask`, `word_tokenize`, `word_pattern_mask`, `aggregate_number`, `aggregate_date`, `scale_number`.
- **`/google_cse`, `/ms_dke`, `/aws` (XKS), `/azureekm`** — already decided (SEC-04, ejected).

## Not audited, but by design (no key operations)

- ui — SPA static files + login
- `/.well-known/jwks.json` — public key export
- `health`, `get_version`, swagger/OpenAPI routes, `root_redirect`

Closes #1067
Closes https://github.com/Cosmian/kms/issues/1064


crate/server/src/middlewares/audit.rs
```diff
@@ -0,0 +1,242 @@
...
+        "AwsXKS".to_owned()
+    } else {
+        // strip leading slash and use first segment
+        path.trim_start_matches('/')
+            .split('/')
+            .next()
+            .unwrap_or("Unknown")
+            .to_owned()
+    }
+}
```
HatemMn: hard coded, take away

crate/server/src/middlewares/audit.rs
```diff
@@ -0,0 +1,242 @@
...
+                    "{} {}",
+                    status.as_u16(),
+                    status.canonical_reason().unwrap_or("Unknown")
+                );
+                make_failure_draft(
+                    timestamp,
+                    final_operation,
+                    final_user,
+                    None,
+                    None,
```
HatemMn: should not be None, to be fixed

HatemMn:
Note: needs 2 reviewers

crate/access/src/audit/cef.rs
```diff
@@ -0,0 +1,271 @@
...
+use time::format_description::well_known::Rfc3339;
+
+use super::event::{AuditEvent, AuditResult};
+
+// CEF severity levels used by this implementation (CEF spec 0.1, header CEF:0)
+const SEV_SUCCESS: u8 = 5;
+const SEV_AUTH_FAILURE: u8 = 7; // 401 / 403
+const SEV_OTHER_FAILURE: u8 = 6;
+
+// Note: no current Rust library currently seems to exist for CEF, hence this implementation is hand-rolled. Tests are pretty heavy.
```
Manuthor, line 12: I did not find crates for CEF format either. But I believe we must ensure compliance with existing CEF format used. Since you may already think about testing, what do you recommend? 

Otherwise, what about this:
Image

crate/access/src/audit/cef.rs
```diff
@@ -0,0 +1,271 @@
...
+        class = escape_header(&event.operation),
+        name = escape_header(&event.operation),
+        sev = severity,
+    );
+
+    // Extension key=value pairs
+    let rt_ms = event.timestamp.unix_timestamp() * 1000 + i64::from(event.timestamp.millisecond());
+
+    let mut ext = format!(
+        "rt={rt} suser={user}",
```
Manuthor, line 57: Could you recall what `rt` stand for? I can't find it in https://www.microfocus.com/documentation/arcsight/arcsight-smartconnectors-8.4/cef-implementation-standard/#CEF/arcsight-extensions.htm?TocPath=_____3.

As a general matter, could you add a 1-line comment to recall the description of each of those CEF fields?

crate/kmip/src/ttlv/kmip_ttlv_serializer.rs
```diff
@@ -31,10 +31,9 @@ impl<T> Stack<T>
 where
     T: Debug,
 {
-    pub(crate) fn new() -> Self {
+    pub(crate) const fn new() -> Self {
         Self {
-            // Pre-allocate for typical KMIP message nesting depth (avoids reallocations)
-            elements: Vec::with_capacity(8),
+            elements: Vec::new(),
```
Manuthor: Please keep this pre-allocation for performance reason (was added on previous release)

For this clarification:

Image

documentation/docs/adr/2026-07-09-003-audit-log-single-writer-design.md
```diff
@@ -0,0 +1,161 @@
...
+title: "ADR-0003: Tamper-Evident JSONL Audit Log — Single-Writer Architecture"
+status: "Accepted"
+date: "2026-07-09"
+authors: "contributors, security architects, compliance engineers"
+tags: ["architecture", "decision", "audit", "compliance", "security"]
+supersedes: ""
+superseded_by: ""
+---
+
+# ADR-0003: Tamper-Evident JSONL Audit Log — Single-Writer Architecture
```
Manuthor, line 11: Can you justify the choice of the JSONL format for the audit file? 

Image

crate/access/src/audit/cef.rs
HatemMn: I did search, when I added that comment I was pretty much sure there is no solid candidate library for Rust (sadly)

To be honest the format is pretty trivial and the heavy hitting unit tests most likely are enough

If I understood this comment correctly (correct me if wrong) you suggest to take some _battle tested_ tool (shall it be in java/python) in order to make some other testing layer ? If yes, this is a very good proposition, I will consider doing this in the next update

HatemMn: Also, in an ideal world, we might want this sort of implementation out of the KMS. For the sake of the project, I propose to do this later.

crate/access/src/audit/cef.rs
HatemMn: Actually all those abreviation have been explained in the table above : check this

https://github.com/Cosmian/kms/blob/7463432f39e925c0656bc0e0a465e19bfde9e8fd/crate/access/src/audit/cef.rs#L24

I might understand that table is not enough docs ? I will try to lean toward a more easier to read doc

by the way, that link is about "ArcSight Extensions" (like, connectors) and not "Arcsight Extension Fields" (without an **S** )

This is the original whitepaper : https://community.opentext.com/cfs-file/__key/telligent-evolution-components-attachments/00-224-01-00-00-15-93-98/CEF-White-Paper-071709.pdf

crate/kmip/src/ttlv/kmip_ttlv_serializer.rs
HatemMn: done, I will make a check I didn't mess this elsewhere

documentation/docs/adr/2026-07-09-003-audit-log-single-writer-design.md
HatemMn: To put it simply:
- JsonL is one of the most cummonly use formats for logs
- the "L" (one entry by line) avoids entirely every issue with formatters and such, while keeping the nested nature of Json that extractors need
- No overhead, No boilerplates, Just logs and some delimiters
- breakages (exp: incomplete line) can be trivially detected
- already ingested by Splunk or even a simple `grep` for indicident response. Also Given that Splunk-UF-on-raw-JSONL is the only production export path we currently have, storing JSONL means the forwarder reads the authoritative file directly. No transform, no second copy to keep in sync. **Any other format will oblige us to either disable splunk export and forward it to a new PR or add a synchro macanism that will make the system slower and error prone**

For the syslog question of the screenshot, yes the standard exists, but it's dead, and the living implementation is proprietary and doesn't do what we'd be doing anyway.

HatemMn (commented):
Sorry for the late posting, I forgot to post this last week

crate/access/src/audit/cef.rs
Manuthor: I believe we should validate now (before touch develop branch) our CEF format against battle tested tool such as the Python format_cef.
Also rt, auditId, timestamp and requestId may not be standard, could we find the appropriate standard fields instead?

crate/access/src/audit/cef.rs
Manuthor: OK thank you. Did you link that paper to the documentation?

documentation/docs/adr/2026-07-09-003-audit-log-single-writer-design.md
Manuthor: And you do confirm that JSONL format is consumable by Splunk, Datadog, etc.?

crate/access/src/audit/cef.rs
HatemMn: As `Python format_cef` does not permet round trips, I used another library for the same pupose (kellyjonbrazil/jc , 8.7k starts on github and maintained), feel free to check this commit : d137a9469a2727ad7bac34dd5541d343e5d3359a

> Also rt, auditId, timestamp and requestId may not be standard, could we find the appropriate standard fields instead?

I think they are https://github.com/Cosmian/kms/pull/934#discussion_r3660522804

crate/access/src/audit/cef.rs
HatemMn: I will add a 1 line comment about it

documentation/docs/adr/2026-07-09-003-audit-log-single-writer-design.md
HatemMn: > And you do confirm that JSONL format is consumable by Splunk, Datadog, etc.?

To my understanding there is a problem with this question, please correct me if I am mixing things:our local file where we store data and "what we push to external systems" are two separate matters.

### First matter
The first is what this PR is about and is _our_ decision that is have justified by these arguments above

### Second matter
The second point is to be done here: https://github.com/Cosmian/kms/issues/881?issue=Cosmian%7Ckms%7C937

Storing our data in JsonL does not have inference to how we will export it to Splunk or friends, here is a doc example : https://github.com/Cosmian/kms/blob/e7bb21ca05a09b99c9769cefa098df4c4db2878d/documentation/docs/kms_clients/audit.md?plain=1#L81

Currently, an **non live** export to those guys is technically possible

[Thanks to a small hack I have documented](https://github.com/Cosmian/kms/blob/e7bb21ca05a09b99c9769cefa098df4c4db2878d/documentation/docs/configuration/siems.md), we can technically start live exporting now, but that's **not yet a feature of the KMS**. It's rather a hack smart operators can pull out - and that I preferred to document 'to expect loss' rather than wait for people to do it and get errors (we can control the KMS, but not what users do with their terminal)

crate/access/src/audit/cef.rs
HatemMn: done

Manuthor (approved):

PR #943, opened 2026-05-06
feat: support PQC X.509 certificates

Manuthor:
#### PQC X.509 Certificates

- **Server — PQC X.509 certificate issuance**: the `Certify` operation now supports ML-DSA-44/65/87 and all SLH-DSA variants (SHA2 / SHAKE × 128s/f / 192s/f / 256s/f) as both subject key algorithms and issuer signing keys (non-FIPS only). The digest selection in `build_and_sign_certificate` previously used the subject's key type and fell through to `SHA-256` for PQC keys; it now correctly uses the issuer's signing key type and maps any non-RSA/EC key (EdDSA, ML-DSA, SLH-DSA) to `MessageDigest::null()` (internal digest).
- **CLI — PQC algorithms in `certificates certify --algorithm`**: the `Algorithm` enum in `certificate_utils` now includes all ML-DSA (`ml-dsa-44`, `ml-dsa-65`, `ml-dsa-87`) and SLH-DSA variants as valid `--algorithm` values for `--generate-key-pair` mode (non-FIPS only).
- **Web UI / WASM — PQC algorithms in Certificate Certify form**: `get_certificate_algorithms()` now includes all ML-DSA and SLH-DSA algorithm options (non-FIPS only), so the *Generate New Keypair* dropdown in the Certificate Certify form exposes all PQC signing algorithms.
- **CLI tests — full PQC algorithm coverage for certificate generation**: added `fetch_pqc_certificate` helper (using `x509_parser` instead of the OpenSSL bindings so PQC OIDs are handled correctly) and `certify_pqc_self_signed` shared helper; added 15 new `#[tokio::test]` cases covering self-signed certificates for every PQC signing algorithm (ML-DSA-65/87, SLH-DSA-SHA2-128s/f, SLH-DSA-SHA2-192s/f, SLH-DSA-SHA2-256s/f, SLH-DSA-SHAKE-128s/f, SLH-DSA-SHAKE-192s/f, SLH-DSA-SHAKE-256s/f) plus a cross-algorithm test (SLH-DSA CA signing an ML-DSA leaf).


.github/scripts/audit/owasp.sh
tbrezot, line 1: Can we require that all files are always formatted with the same tool (checked by the CI) so as to avoid diffs like that in reviews ?

.github/scripts/test/test_hsm_crypt2pay.sh
```diff
@@ -77,7 +77,7 @@ if ! source "$REPO_ROOT/.github/reusable_scripts/prepare_crypt2pay.sh"; then
     echo "Continuing: Crypt2Pay client appears installed despite prepare script self-test failure."
   else
     echo "Error: Crypt2Pay client setup is incomplete."
-    exit 1
+    exit 0
```
tbrezot: Why is this change necessary ? Isn't returning a non-zero value widely used as a sign of failure ?

crate/clients/client_utils/src/certificate_utils.rs
tbrezot, line 1: Aren't some of those algorithms FIPS-compliant ?

crate/server/src/core/operations/certify/mod.rs
```diff
@@ -587,7 +594,31 @@ async fn issuer_for_self_signed_certificate<'a>(
...
+                    kms_bail!(KmsError::InvalidRequest(
+                        "KEM keys (ML-KEM, X25519MLKEM768, X448MLKEM1024, ConfigurableKEM) \
+                         do not support digital signatures and cannot issue self-signed \
+                         certificates. Provide issuer_private_key_id and \
+                         issuer_certificate_id pointing to a signing CA \
+                         (RSA/EC/ML-DSA/SLH-DSA)."
+                            .to_owned()
+                    ))
+                }
+            }
```
tbrezot: Instead of listing the keys that cannot be used, list the keys that can, and returns this error in case the given key is not one of them. E.g. what happens if we never enter the first if (the object is a private key) ? Is this a valid branch ?

Also -- this is a detail -- `pk` is reserved for public keys. Prefer `sk` for private keys.

crate/server/src/core/operations/validate.rs
```diff
@@ -225,21 +226,38 @@ fn sort_certificates(certificates: &[X509]) -> KResult<Vec<X509>> {
...
+            .is_some_and(|(subj, iss)| !subj.is_empty() && subj == iss);
+
+        if (is_root_explicit || is_root_self_signed) && !sorted_chains.contains(certificate) {
             trace_certificate("Root found", certificate);
             sorted_chains.insert(0, certificate.to_owned());
             indexes_to_remove.push(index);
-        }
-
-        if aki.is_empty() && ski.is_empty() {
+        } else if aki.is_empty() && ski.is_empty() && !sorted_chains.contains(certificate) {
```
tbrezot: There is something fishy about this duplicated `!sorted_chains.contains(certificate)` check, are you sure about your logic here?

crate/server/src/core/operations/certify/mod.rs
```diff
@@ -672,22 +817,62 @@ fn build_and_sign_certificate(
...
+    // The extension value per RFC 9608 §4.3.1 is NULL (DER: 05 00).
+    let is_self_signed = matches!(issuer, Issuer::PrivateKeyAndSubjectName(..));
+    if is_self_signed && !ext_conf_has_cdp {
+        let oid = Asn1Object::from_str("1.3.6.1.5.5.7.1.56")?;
+        let val = Asn1OctetString::new_from_bytes(&[0x05, 0x00])?;
+        x509_builder.append_extension(X509Extension::new_from_der(
+            oid.as_ref(),
+            false,
+            val.as_ref(),
+        )?)?;
```
tbrezot: I cannot understand  what is happening here:
- I couldn't find any mention of `ip-de-noRevAvail` in RFC9608 section 4, but I could find `id-ce-noRevAvail` in [RFC9608 section 2](https://www.rfc-editor.org/rfc/rfc9608.html#section-2)
- I could not find any mention of the OID used here, but I could find the OID prefix `2.5.29` associated to `id-ce` in [RFC5280 section 4.2.1](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1)
- I could not find the section 4.3.1 from RFC 9608.

Could you please provide more information ? If those lines are written with the help of an AI could you explicitly state it in the MR message and explain what procedure was used to assert their correctness? For example is there a test of compatibility with an external X509 certificate tool ?

tbrezot (commented):
The code is very hard to read because implementation of a given RFC is split across different functions and interleaved with the implementation of other RFCs. I understand this is tricky code to write, but do you think it would be possible to have a single function implementing a single RFC? This would lead to a much more readable code imho.

.github/scripts/test/test_hsm_crypt2pay.sh
Manuthor: I can revert it but all the CI pipelines will be red from time to time (since Crypt2Pay HSM can be down)

crate/clients/client_utils/src/certificate_utils.rs
Manuthor: The algorithms are FIPS-compliant but last OpenSSL FIPS provider remains the version 3.1.2

.github/scripts/audit/owasp.sh
Manuthor: While not changing the CI, I've added a pre-commit local hook calling shfmt

crate/server/src/core/operations/certify/mod.rs
Manuthor: Done in cc2a9628fcd164918e3e4ada296ec9812697b073

crate/server/src/core/operations/validate.rs
Manuthor: Right, thank you. It should be okay in cc2a9628fcd164918e3e4ada296ec9812697b073

crate/server/src/core/operations/certify/mod.rs
Manuthor: Completely right...
Should be okay now (cc2a9628fcd164918e3e4ada296ec9812697b073)

Issue #861, opened 2026-03-29
Kubernetes KMS Provider Plugin (etcd Secret encryption)


Manuthor:
### Problem

Kubernetes can encrypt Secrets at rest in etcd via the [KMS Provider Plugin v2 API](https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/). No Cosmian-KMS plugin exists today.

### Proposed solution

New standalone binary crate `crate/clients/k8s_plugin/` that implements the gRPC KMS v2 Unix-socket server and delegates encryption / decryption to `cosmian_kms_client`.

#### Architecture

```
kube-apiserver
  │  (Unix socket: /var/run/cosmian-kms-plugin.sock)
  ▼
cosmian-kms-plugin  (new binary crate)
  │  (HTTPS + mTLS)
  ▼
Cosmian KMS server
```

#### gRPC interface (kmsv2.proto — bundle in crate)

| RPC | Description |
|---|---|
| `Status()` | Health-check; return KMS version |
| `Encrypt()` | AES-256-GCM wrap of DEK (delegate to `KmsClient::encrypt`) |
| `Decrypt()` | Unwrap DEK (delegate to `KmsClient::decrypt`) |

#### Config (YAML — Kubernetes EncryptionConfiguration compatible)

```yaml
cosmian_kms:
  server_url: https://kms.example.com:9998
  api_key: ""            # or use client cert below
  tls_cert: /etc/cosmian-kms-plugin/client.crt
  tls_key: /etc/cosmian-kms-plugin/client.key
  ca_cert: /etc/cosmian-kms-plugin/ca.crt
  wrapping_key_uid: "..."
  socket_path: /var/run/cosmian-kms-plugin.sock
```

No changes to the KMS server are required — the plugin uses the existing `cosmian_kms_client`.

### Acceptance criteria

- A `kind` cluster configured with the matching `EncryptionConfiguration` successfully stores and retrieves Kubernetes Secrets.
- Raw etcd values are opaque (AES-256-GCM encrypted) when inspected with `etcdctl`.
- Plugin restarts without data loss (DEK is re-wrapped on each `Encrypt`; plaintext DEK is never persisted).
- `Status()` is polled by kube-apiserver every 60 s; plugin responds within 5 s or is marked degraded.

### Files to create

| File | Change |
|---|---|
| `crate/clients/k8s_plugin/Cargo.toml` | **New** — binary crate |
[...]

PR #950, opened 2026-05-12
ci: add Kubernetes KMS v2 HSM plugin integration tests (#861)

p0wline:


Manuthor (changes requested):
If you do not mind, I would rather move this CI directly on https://github.com/Cosmian/k8s-hsm-kmsv2.git (since KMS CI is already hard and IMHO it is more appropriate on https://github.com/Cosmian/k8s-hsm-kmsv2.git)

PR #968, opened 2026-05-28
feat: generalize objects rotation - all key types + certificates

Manuthor:
## Summary

Delivers the complete **key auto-rotation specification** and the full manual-rotation
implementation for all key types, as PR 1 of 3 in the auto-rotation feature stack.

```
develop
  ← PR 1  docs/key-autorotation-spec          ← this PR (#968)
  ← PR 2  feat/key-rotation-scheduler          (#970)
  ← PR 3  feat/key-rotation-notifications      (#971)
```

---

## Documentation

📄 **[Key Auto-Rotation Policy](https://docs.cosmian.com/key_management_system/kmip_support/key_auto_rotation/)** — published after merge.
Source: [`documentation/docs/kmip_support/key_auto_rotation.md`](https://github.com/Cosmian/kms/blob/docs/key-autorotation-spec/documentation/docs/kmip_support/key_auto_rotation.md)

### What the doc covers

- **Rotation-policy attributes** — six vendor-extension KMIP attributes (`x-rotate-interval`, `x-rotate-name`, `x-rotate-offset`, `x-rotate-generation`, `x-rotate-date`, `x-rotate-latest`). `x-rotate-generation` and `x-rotate-date` are server-managed and read-only to prevent chain corruption.
- **Keysets** — named groups of successive key generations. Supports bare-name (`my-keyset`) and `my-keyset@N` / `my-keyset@latest` addressing. Encrypt/Sign always use the latest generation; Decrypt/Verify walk newest-to-oldest so in-flight ciphertexts survive rotation.
- **HSM key rotation** — HSM-resident keys (`hsm::…` UIDs) support manual `Re-Key` (calls `C_GenerateKey` on the same PKCS#11 slot) and keysets (state stored in `CKA_LABEL`), but are **excluded from the auto-rotation scheduler**.
- **Server-side scheduler** — background cron thread, enabled via `--auto-rotation-check-interval-secs`. Fires `Re-Key` automatically when `x-rotate-date + x-rotate-interval` elapses.
- **State restrictions** — Active, Deactivated, and Compromised keys may be rotated; Pre-Active / Destroyed are rejected.
- **Key types and rotation flows** — seven scenarios with Mermaid diagrams: plain symmetric key, wrapping key (re-wraps all dependants in a two-phase atomic commit), wrapped key, asymmetric key pair, wrapped private key / CoverCrypt, certificate renewal (`ReCertify`), server-wide KEK.

---

## Implementation — `Re-Key`, `Re-Key Key Pair`, `Re-Certify`

Complete manual-rotation implementation for all six scenarios:

1. **Plain symmetric key** — new material, new UID, rotation metadata
2. **Wrapping key** — rotate + re-wrap all dependants atomically
3. **Wrapped key** — unwrap → new material → re-wrap
4. **Asymmetric key pair** — new private + public UIDs, all curve types (RSA, EC, ML-KEM, ML-DSA, SLH-DSA, X25519, secp256k1, CoverCrypt)
5. **Wrapped private key / CoverCrypt** — CoverCrypt attribute-level rekey with wrapping preserved
6. **Server-wide KEK** — transparent via default wrapping config

Implements `ReCertify` (KMIP 2.1 §6.1.45): self-signed and CA-signed renewal, replacement link chain, offset-based `PreActive` state.

### Code quality

- All standalone functions dispatched to appropriate struct `impl` blocks — rotation helpers moved onto `impl Attributes`, `impl Object`, `impl UniqueIdentifier` in `crate/kmip`; crypto-op helpers moved onto `impl KMS` in `crate/server`.
- Zero Clippy warnings. 344 tests pass (201 FIPS, 331 non-FIPS).

---

## Breaking changes

None.

---

## Reviewer notes

This PR is the canonical reference for all subsequent PRs in this stack.
Please review terminology and attribute semantics carefully — changes here cascade to the implementation PRs.


crate/server/src/core/operations/rekey/common.rs
```diff
@@ -0,0 +1,852 @@
+//! Shared logic for KMIP `ReKey` (§4.4), `ReKeyKeyPair` (§4.5), and `ReCertify` (§4.7) operations.
```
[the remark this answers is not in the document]
Manuthor: Fixed: updated the module header to reference ReCertify as KMIP §4.8 (KMIP 1.4) / §6.1.45 (KMIP 2.1).

crate/server/src/core/operations/rekey/common.rs
```diff
@@ -0,0 +1,852 @@
...
+
+    for (dep_uid, _dep_state, dep_attrs) in wrapped_dependants {
+        let Some(dep_owm) = kms.database.retrieve_object(&dep_uid).await? else {
+            warn!("wrapped dependant {dep_uid} not found, skipping");
+            continue;
+        };
+        let mut dep_object = dep_owm.object().clone();
+
+        if let Some(op) =
+            rewrap_single_dependant(kms, owner, &dep_uid, &mut dep_object, dep_attrs, new_uid)
```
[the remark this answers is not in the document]
Manuthor: Fixed: added an ownership check in `rewrap_dependants` — dependants where `dep_owm.owner() != owner` are now skipped with a warning log.

crate/kmip/src/kmip_2_1/kmip_operations.rs
```diff
@@ -1021,6 +1027,66 @@ pub struct CertifyResponse {
...
+/// The request contains the Unique Identifier of the existing certificate to be
+/// renewed, an optional certificate request, and optional attributes for the new
+/// certificate.
+///
+/// The server creates a new Certificate object with a fresh Unique Identifier,
+/// sets a `ReplacedObjectLink` on the new certificate pointing to the old one,
+/// and sets a `ReplacementObjectLink` on the old certificate pointing to the new one.
+///
+/// KMIP 2.1 §6.1.8 / KMIP 1.4 §4.8
+#[derive(Clone, Default, Serialize, Deserialize, PartialEq, Eq, Debug)]
```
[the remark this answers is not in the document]
Manuthor: Fixed: updated the doc comment to reference §6.1.45 (KMIP 2.1).

crate/server/src/core/operations/recertify.rs
```diff
@@ -0,0 +1,347 @@
...
+    old_cert_attrs: &Attributes,
+    new_cert_uid: &str,
+    operations: &mut Vec<AtomicOperation>,
+) -> KResult<()> {
+    let old_cert_uid = old_cert_attrs
+        .unique_identifier
+        .as_ref()
+        .map(std::string::ToString::to_string)
+        .unwrap_or_default();
+
```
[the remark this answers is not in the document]
Manuthor: Fixed: `relink_keys_to_new_certificate` now takes `old_cert_uid: &str` as a parameter (the actual object ID from the DB) instead of deriving it from attributes.

crate/server_database/src/stores/sql/sqlite.rs
```diff
@@ -531,6 +531,63 @@ impl ObjectsStore for SqlitePool {
...
+             FROM objects \
+             LEFT JOIN read_access ON objects.id = read_access.id \
+                 AND read_access.userid = $2 \
+             WHERE (objects.owner = $2 OR read_access.userid = $2) \
+               AND ( \
+                 json_extract(objects.object, '$.SymmetricKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier') = $1 \
+                 OR json_extract(objects.object, '$.PrivateKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier') = $1 \
+                 OR json_extract(objects.object, '$.SecretData.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier') = $1 \
+                 OR json_extract(objects.object, '$.SplitKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier') = $1 \
+                 OR json_extract(objects.object, '$.PGPKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier') = $1 \
```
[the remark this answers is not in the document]
Manuthor: Fixed: added the `PublicKey` variant to the SQLite `find_wrapped_by` inline query.

crate/server_database/src/stores/sql/pgsql.rs
```diff
@@ -780,6 +780,46 @@ impl ObjectsStore for PgPool {
...
+                LEFT JOIN read_access ON objects.id = read_access.id \
+                    AND read_access.userid = $2 \
+                WHERE (objects.owner = $2 OR read_access.userid = $2) \
+                  AND ( \
+                    objects.object->'SymmetricKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'PrivateKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'SecretData'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'SplitKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'PGPKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                  )";
```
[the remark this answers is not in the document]
Manuthor: Fixed: added the `PublicKey` JSON path to the PostgreSQL `find_wrapped_by` query in `query.sql`.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -625,6 +625,57 @@ impl ObjectsStore for MySqlPool {
...
+            LEFT JOIN read_access ON objects.id = read_access.id \
+                AND read_access.userid = ? \
+            WHERE (objects.owner = ? OR read_access.userid = ?) \
+              AND ( \
+                objects.object->>'$.SymmetricKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.PrivateKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.SecretData.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.SplitKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.PGPKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+              )";
```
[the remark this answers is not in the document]
Manuthor: Fixed: added the `PublicKey` JSON path to the MySQL `find_wrapped_by` query in `query_mysql.sql`, along with the corresponding bind parameter.

crate/server/src/core/operations/rekey/common.rs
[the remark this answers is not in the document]
Manuthor: Fixed: updated the module header to reference ReCertify as KMIP §4.8 (KMIP 1.4) / §6.1.45 (KMIP 2.1).

crate/server/src/core/operations/rekey/common.rs
```diff
@@ -0,0 +1,852 @@
...
+        .find_wrapped_by(old_uid, owner)
+        .await
+        .unwrap_or_default();
+
+    for (dep_uid, _dep_state, dep_attrs) in wrapped_dependants {
+        let Some(dep_owm) = kms.database.retrieve_object(&dep_uid).await? else {
+            warn!("wrapped dependant {dep_uid} not found, skipping");
+            continue;
+        };
+        let mut dep_object = dep_owm.object().clone();
```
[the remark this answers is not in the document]
Manuthor: Fixed: added an ownership check in `rewrap_dependants` — dependants where `dep_owm.owner() != owner` are now skipped with a warning log.

crate/kmip/src/kmip_2_1/kmip_operations.rs
```diff
@@ -1021,6 +1027,66 @@ pub struct CertifyResponse {
...
+/// existing public key whose certificate has expired or is about to expire.
+/// The request contains the Unique Identifier of the existing certificate to be
+/// renewed, an optional certificate request, and optional attributes for the new
+/// certificate.
+///
+/// The server creates a new Certificate object with a fresh Unique Identifier,
+/// sets a `ReplacedObjectLink` on the new certificate pointing to the old one,
+/// and sets a `ReplacementObjectLink` on the old certificate pointing to the new one.
+///
+/// KMIP 2.1 §6.1.8 / KMIP 1.4 §4.8
```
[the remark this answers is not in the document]
Manuthor: Fixed: updated the doc comment to reference §6.1.45 (KMIP 2.1).

crate/server/src/core/operations/recertify.rs
```diff
@@ -0,0 +1,347 @@
...
+    _user: &str,
+    old_cert_attrs: &Attributes,
+    new_cert_uid: &str,
+    operations: &mut Vec<AtomicOperation>,
+) -> KResult<()> {
+    let old_cert_uid = old_cert_attrs
+        .unique_identifier
+        .as_ref()
+        .map(std::string::ToString::to_string)
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
Manuthor: Fixed: `relink_keys_to_new_certificate` now takes `old_cert_uid: &str` as a parameter (the actual object ID from the DB) instead of deriving it from attributes.

crate/server_database/src/stores/sql/sqlite.rs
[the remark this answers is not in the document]
Manuthor: Fixed: added the `PublicKey` variant to the SQLite `find_wrapped_by` inline query.

crate/server_database/src/stores/sql/pgsql.rs
```diff
@@ -780,6 +780,46 @@ impl ObjectsStore for PgPool {
...
+                FROM objects \
+                LEFT JOIN read_access ON objects.id = read_access.id \
+                    AND read_access.userid = $2 \
+                WHERE (objects.owner = $2 OR read_access.userid = $2) \
+                  AND ( \
+                    objects.object->'SymmetricKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'PrivateKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'SecretData'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'SplitKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
+                    OR objects.object->'PGPKey'->'KeyBlock'->'KeyWrappingData'->'EncryptionKeyInformation'->>'UniqueIdentifier' = $1 \
```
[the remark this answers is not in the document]
Manuthor: Fixed: added the `PublicKey` JSON path to the PostgreSQL `find_wrapped_by` query in `query.sql`.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -625,6 +625,57 @@ impl ObjectsStore for MySqlPool {
...
+            FROM objects \
+            LEFT JOIN read_access ON objects.id = read_access.id \
+                AND read_access.userid = ? \
+            WHERE (objects.owner = ? OR read_access.userid = ?) \
+              AND ( \
+                objects.object->>'$.SymmetricKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.PrivateKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.SecretData.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.SplitKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
+                OR objects.object->>'$.PGPKey.KeyBlock.KeyWrappingData.EncryptionKeyInformation.UniqueIdentifier' = ? \
```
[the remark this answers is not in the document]
Manuthor: Fixed: added the `PublicKey` JSON path to the MySQL `find_wrapped_by` query in `query_mysql.sql`, along with the corresponding bind parameter.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -625,6 +625,57 @@ impl ObjectsStore for MySqlPool {
...
+                (
+                    user,
+                    user,
+                    user,
+                    wrapping_key_uid,
+                    wrapping_key_uid,
+                    wrapping_key_uid,
+                    wrapping_key_uid,
+                    wrapping_key_uid,
+                ),
```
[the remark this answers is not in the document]
Manuthor: Fixed: the `query_mysql.sql` now includes the extra `?` placeholder for `PublicKey` and the exec parameter tuple has the additional `wrapping_key_uid` binding.

crate/server_database/src/stores/sql/mysql.rs
```diff
@@ -625,6 +625,57 @@ impl ObjectsStore for MySqlPool {
...
+            )
```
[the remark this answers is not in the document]
Manuthor: Fixed: the `query_mysql.sql` now includes the extra `?` placeholder for `PublicKey` and the exec parameter tuple has the additional `wrapping_key_uid` binding.

crate/interfaces/src/stores/objects_store.rs
```diff
@@ -102,4 +102,19 @@ pub trait ObjectsStore {
         user_must_be_owner: bool,
         vendor_id: &str,
     ) -> InterfaceResult<Vec<(String, State, Attributes)>>;
+
+    /// Return (uid, state, attributes) for every object whose
```
tbrezot, line 108: A note for the future (will raise this comment in a dedicated issue if you agree): UIDs are not typed. However, we have different kinds of IDs: User IDs, Key IDs... all are strings. The code-base would be improved if those were of distinct types.

crate/interfaces/src/stores/objects_store.rs
```diff
@@ -102,4 +102,19 @@ pub trait ObjectsStore {
...
+    /// `key_wrapping_data.encryption_key_information.unique_identifier` equals
+    /// `wrapping_key_uid`. Used by key rotation to re-wrap all objects protected by
+    /// the rotated key.
+    ///
+    /// The default implementation returns an empty list; backends that support
+    /// JSON-based object storage should override this with an efficient query.
```
tbrezot: It seems to me this function can be implemented by every SQL back-end by simply adding a column to the schema. I guess NoSQL database could also implement it efficiently.

Does this mean that when using a back-end that does not support JSON-based object storage it is not possible to use endpoints using this function?

crate/kmip/src/kmip_1_4/kmip_operations.rs
```diff
@@ -498,6 +498,60 @@ pub struct ReCertifyResponse {
...
+        let cert_req_type = match recertify.certificate_request_type {
+            Some(kmip_2_1::kmip_types::CertificateRequestType::CRMF) => {
+                CertificateRequestType::CRMF
+            }
+            Some(kmip_2_1::kmip_types::CertificateRequestType::PKCS10) => {
+                CertificateRequestType::PKCS10
+            }
+            Some(kmip_2_1::kmip_types::CertificateRequestType::PEM) | None => {
+                CertificateRequestType::PEM
+            }
```
tbrezot: Why is the `None` value converted to the `PEM` variant? Is this documented?

crate/kmip/src/kmip_1_4/kmip_operations.rs
```diff
@@ -498,6 +498,60 @@ pub struct ReCertifyResponse {
...
+        };
+        Self {
+            unique_identifier: recertify
+                .unique_identifier
+                .map_or_else(String::new, |u| u.to_string()),
```
tbrezot: Isn't the lack of ID equivalent to using the "ID Placeholder"?

crate/kmip/src/kmip_2_1/kmip_operations.rs
```diff
@@ -1021,6 +1027,66 @@ pub struct CertifyResponse {
...
+/// and sets a `ReplacementObjectLink` on the old certificate pointing to the new one.
+///
+/// KMIP 2.1 §6.1.45 / KMIP 1.4 §4.8
+#[derive(Clone, Default, Serialize, Deserialize, PartialEq, Eq, Debug)]
+#[serde(rename_all = "PascalCase")]
+pub struct ReCertify {
+    /// The Unique Identifier of the existing Certificate to be re-certified.
+    /// If omitted, the ID Placeholder value is used.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub unique_identifier: Option<UniqueIdentifier>,
```
tbrezot: Isn't it possible to specify here a default value for serde? Otherwise this invariant is non-local and thus hard to verify.

crate/server/src/core/operations/key_ops/mod.rs
```diff
@@ -54,6 +54,9 @@ pub(crate) fn setup_object_lifecycle(
     attributes.last_change_date = Some(now);
     if state == State::Active {
         attributes.activation_date = Some(now);
+    } else if let Some(future_date) = requested_activation_date {
+        // PreActive: store the future activation date so auto-transition works
+        attributes.activation_date = Some(future_date);
```
tbrezot: It would be better to fail in case there is a requested activation date but the object is already active rather than silently ignoring it.

crate/server/src/core/wrapping/wrap.rs
```diff
@@ -69,11 +69,21 @@ pub(crate) async fn wrap_and_cache(
...
+            return Err(KmsError::InvalidRequest(format!(
+                "Key '{wrapping_key_id}' cannot be used as its own wrapping key: \
+                 the wrapping key ID must differ from the key ID being created"
+            )));
         }
+        // The server-wide KEK coincidentally matches the new key's UID — skip silently.
+        // This should not happen in practice (KEKs use prefixed UIDs like "hsm::softhsm2::0::kek").
+        warn!(
+            "Server KEK '{wrapping_key_id}' matches the UID of the key being created; skipping self-wrap"
+        );
```
tbrezot: I do not know why we do not fail *all the time* since whether or not it was already specified in the object, wrapping itself is an illegal operation.

crate/server/src/core/wrapping/wrap.rs
```diff
@@ -245,7 +255,15 @@ async fn wrap_using_kms(
...
-    if wrapping_key.owner() != user {
+    // The server-configured key_encryption_key is a shared server resource accessible
+    // to all users, so skip the ownership check for it (mirrors the bypass in
+    // `wrap_using_crypto_oracle` — issue #761).
+    let is_server_kek = kms
+        .params
+        .key_wrapping_key
+        .as_deref()
+        .is_some_and(|kek| kek == wrapping_key_uid);
+    if !is_server_kek && wrapping_key.owner() != user {
```
tbrezot, line 266: Why can't a user that has encryption right over a given key cannot use it to wrap its keys? Thinking about it maybe because this right is revocable while the ownership isn't?

crate/server/src/core/kms/kmip.rs
```diff
@@ -669,6 +669,24 @@ impl KMS {
...
+    /// `ReCertify` — certificate rotation with a new UID.
+    ///
+    /// Creates a fresh certificate for the same subject/issuer and links old → new
+    /// via `ReplacementObjectLink`. Keys referencing the old certificate are updated
+    /// to point to the new one.
+    pub(crate) async fn recertify(
+        &self,
+        request: ReCertify,
+        user: &str,
+        privileged_users: Option<Vec<String>>,
```
tbrezot: Thinking about it: why isn't `privileged_users` in `self`? If I understand it correctly this is a global parameter.

crate/server_database/src/stores/redis/redis_with_findex.rs
```diff
@@ -622,6 +622,58 @@ impl ObjectsStore for RedisWithFindex {
...
+                    .object
+                    .attributes()
+                    .cloned()
+                    .unwrap_or_else(|_| Attributes {
+                        object_type: Some(dbo.object.object_type()),
+                        ..Default::default()
+                    });
+                out.push((uid, dbo.state, attrs));
+            }
+        }
```
tbrezot: This seems to be rather inefficient. Isn't it possible to simply index wrapped key under the wrapping-key ID?

crate/server_database/src/stores/sql/query.sql
tbrezot, line 1: This query could be optimized by adding a column to index the wrapping-key ID.

crate/server/src/core/operations/rekey/symmetric.rs
```diff
@@ -0,0 +1,194 @@
...
+    set_rotation_metadata_on_new_key, validate_no_crypto_param_change,
+};
+use crate::{
+    core::{KMS, operations::key_ops::ObjectWithMetadataOps},
+    error::KmsError,
+    kms_bail,
+    result::{KResult, KResultHelper},
+};
+
+/// Implementor of [`RekeyOperation`] for KMIP `ReKey` (§4.4) on symmetric keys.
```
tbrezot: Section 4.4 of KMIP specification 2.1 is about "Application Specific Information", not re-key. Is this an hallucinated reference? Can you prevent your LLM from inserting references as almost all of them are incorrect?

crate/server/src/core/operations/rekey/symmetric.rs
```diff
@@ -0,0 +1,194 @@
...
+pub(crate) struct SymmetricRekey {
+    /// The `offset` from the `ReKey` request (date arithmetic per KMIP Table 172).
```
tbrezot: Table 172 of KMIP 2.1 is about "Adjust Attribute Errors" not offset or "date arithmetic". The term date-arithmetic is nowhere to be found in the norm. What is the meaning of it? Is an `i32` really enough to hold it? The only information I could find in the norm is that it is added to -- and should therefore be homogeneous with -- an activation date which is in a `Date-Time` format that seems to be left unspecified.

crate/server/src/core/operations/rekey/symmetric.rs
```diff
@@ -0,0 +1,194 @@
...
+    offset: Option<i32>,
+}
+
+/// KMIP `ReKey` operation for symmetric keys.
+///
+/// Per KMIP 1.4 §4.4:
```
tbrezot: Why is KMIP 1.4 cited here? I though the KMS was implementing the v2.1 norm and converted request (resp. responses) from (resp to) the v1.4 format?

crate/server/src/core/operations/rekey/symmetric.rs
```diff
@@ -0,0 +1,194 @@
...
+    request: ReKey,
+    owner: &str,
+    privileged_users: Option<Vec<String>>,
+) -> KResult<ReKeyResponse> {
+    trace!("ReKey: {}", serde_json::to_string(&request)?);
+    let offset = request.offset;
+    execute_rekey(
+        &SymmetricRekey { offset },
+        kms,
+        &request,
```
tbrezot: The offset is passed twice here: in the request and in a dedicated structure. Maybe consider passing it only once.

crate/server/src/core/operations/rekey/common.rs
```diff
@@ -0,0 +1,864 @@
...
+/// Returns `None` if the object has no key block or is not wrapped.
+/// Used by the default [`RekeyOperation::detect_wrapping`] implementation.
+pub(crate) fn extract_rewrap_spec(object: &Object) -> Option<KeyWrappingSpecification> {
+    let kb = object.key_block().ok()?;
+    let kwd = kb.key_wrapping_data.as_ref()?;
+    Some(KeyWrappingSpecification {
+        wrapping_method: kwd.wrapping_method,
+        encryption_key_information: kwd.encryption_key_information.clone(),
+        mac_or_signature_key_information: kwd.mac_signature_key_information.clone(),
+        attribute_name: None,
```
tbrezot: Why is this none ? Shouldn't it be extractable from an `object` ?

crate/server/src/core/operations/rekey/common.rs
```diff
@@ -0,0 +1,864 @@
...
+/// Returns `None` if the object is not wrapped or has no `EncryptionKeyInformation`.
+pub(crate) fn extract_wrapping_key_uid(object: &Object) -> Option<String> {
+    object
+        .key_block()
+        .ok()
+        .and_then(|kb| kb.key_wrapping_data.as_ref())
+        .and_then(|kwd| kwd.encryption_key_information.as_ref())
+        .and_then(|eki| eki.unique_identifier.as_str())
+        .map(str::to_owned)
+}
```
tbrezot: To me there are two cases here: either there is no wrapping information and the function should return `None`, either there is wrapping information but the wrapping-key ID is missing and this is a violation of the structural invariants of the KMS objects and this should be an error.

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+
+## Rotation policy attributes
+
+All rotation-policy state is stored as vendor-extension KMIP attributes on
+the key object itself.  The following attributes are available:
+
+| Attribute             | Type            | Description                                                                  |
+| --------------------- | --------------- | ---------------------------------------------------------------------------- |
+| `x-rotate-interval`   | `u32` (seconds) | How often this key should be rotated. `0` disables auto-rotation.            |
+| `x-rotate-name`       | `String`        | Optional human-readable label for the policy (e.g. `"daily"`, `"annual"`).   |
```
tbrezot: Consider renaming this variable into `x-rotate-policy` and to use an enumeration to represent it internally.

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+| `x-rotate-offset`     | `u32` (seconds) | Shift the first rotation trigger by this many seconds after `Initial Date`.  |
+| `x-rotate-generation` | `u64`           | Incremented on every rotation; `0` for never-rotated keys.                   |
```
tbrezot: A user should not be able to modify this attribute as I guess some functions of the KMS rely it to preserve some invariant (e.g. that it is unique given a key-set, or that it is strictly increasing).

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+| `x-rotate-date`       | `datetime`      | Timestamp of the last rotation; populated automatically after each rotation. |
```
tbrezot: Isn't there an object creation date that could replace this parameter everywhere? (it seems natural to keep this kind of information around).

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+---
+
+## Rotation policy attributes
+
+All rotation-policy state is stored as vendor-extension KMIP attributes on
+the key object itself.  The following attributes are available:
+
+| Attribute             | Type            | Description                                                                  |
+| --------------------- | --------------- | ---------------------------------------------------------------------------- |
+| `x-rotate-interval`   | `u32` (seconds) | How often this key should be rotated. `0` disables auto-rotation.            |
```
tbrezot: This type does not correspond to the implementation (`i32`).

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+| `x-rotate-name`       | `String`        | Optional human-readable label for the policy (e.g. `"daily"`, `"annual"`).   |
+| `x-rotate-offset`     | `u32` (seconds) | Shift the first rotation trigger by this many seconds after `Initial Date`.  |
```
tbrezot: This type does not corresponds to the implementation (`i32`).

documentation/docs/kmip_support/key_auto_rotation.md
tbrezot: This type does not correspond to the implementation (`i32`)

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+    --name    "hourly"
+```
+
+---
+
+> **⚠️ HSM-resident keys cannot be auto-rotated**
+>
+> Keys whose UID starts with `hsm::` are stored entirely inside the Hardware
+> Security Module. The KMS has no ability to generate new key material inside
+> the HSM, replace an existing HSM key, or migrate key material to a new UID.
```
tbrezot: Is it really true that the KMS cannot require the creation of a new key in the HSM? I can see a method `create_key` in the `HSM` trait:
```
    /// Create the given key in the HSM.
    /// The key ID will be generated by the HSM and returned.
    ///
    /// The key will not be exportable from the HSM if the sensitive flag is set to true.
    /// # Arguments
    /// * `slot_id` - the slot ID of the HSM
    /// * `id` - the ID of the key
    /// * `algorithm` - the key algorithm to use
    /// * `key_length_in_bits` - the length of the key in bits
    /// * `sensitive` - whether the key should be exportable
    /// # Returns
    /// * `PluginResult<usize>` - the ID of the key
    async fn create_key(
        &self,
        slot_id: usize,
        id: &[u8],
        algorithm: HsmKeyAlgorithm,
        key_length_in_bits: usize,
        sensitive: bool,
    ) -> InterfaceResult<()>;
```

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+> As a result:
+>
+> - `find_due_for_rotation` never returns HSM UIDs (they are not in the KMS
+>   database), so the scheduler will never attempt to rotate them.
+> - Calling `Re-Key` manually on an `hsm::` UID will fail.
+> - Setting `x-rotate-interval` on an HSM key is unsupported and has no effect.
```
tbrezot: Maybe raise an error instead?

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+
+---
+
+### 1. Plain symmetric key (no wrapping)
+
+A plain symmetric key carries only its own policy.  On rotation:
+
+1. Fresh key material is generated (same algorithm and length).
+2. The new key is assigned a new UUID.
+3. A `ReplacedObjectLink` on the new key points back to the old key.
```
tbrezot: What is the convention when there is no old key (this is the first key of the set)?

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+4. A `ReplacementObjectLink` on the old key points forward to the new key.
```
tbrezot: What is the convention when this is the latest key of the set?

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+A **CoverCrypt** private key that has been wrapped follows the same flow as any
+other `PrivateKey` rotation: the `ReKeyKeyPair` (`rekey_keypair`) operation
+unwraps the key in memory, rekeys the CoverCrypt partition, and stores a new
+wrapped private key under a fresh UID.
+
+> **Note on RSA / EC private keys**: auto-rotation of RSA and EC private keys
+> via `ReKeyKeyPair` is not yet supported.  If a rotation policy is set on an
+> RSA or EC private key, the scheduler will attempt rotation and log a warning
+> instead of failing.
+
```
tbrezot: Why are we talking about RSA/EC in the Covercrypt section ? Plus aren't RSA/EC private keys asymmetric ? Why don't they fall into the previous section?

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+```
+
+---
+
+### 5. Wrapped private key (CoverCrypt)
+
+A **CoverCrypt** private key that has been wrapped follows the same flow as any
+other `PrivateKey` rotation: the `ReKeyKeyPair` (`rekey_keypair`) operation
+unwraps the key in memory, rekeys the CoverCrypt partition, and stores a new
+wrapped private key under a fresh UID.
```
tbrezot: By "Covercrypt private key" do you mean "Covercrypt master private key" ? What about the user keys ?

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+```
+
+---
+
+### 6. Certificate renewal (`ReCertify`)
+
+Certificate renewal creates a **new certificate for the same key pair** — no
+new key material is generated.  The KMIP `ReCertify` operation (§6.1.45)
+assigns a fresh UID to the renewed certificate and links old → new via the
+standard `ReplacementObjectLink` / `ReplacedObjectLink` pair.
```
tbrezot: Are we generating new certificates for rotated asymmetric keys?

documentation/docs/kmip_support/key_auto_rotation.md
```diff
@@ -0,0 +1,623 @@
...
+
+# Self-signed certificate renewal (issuer = subject)
+ckms certificates certify \
+    --certificate-id-to-re-certify <OLD_CERT_UID> \
+    --days 3650
+```
+
+---
+
+### 7. Server-wide key-encryption key (KEK)
```
tbrezot: What happens if we want to rotate the master KEK? Is it currently supported?

documentation/docs/kmip_support/key_auto_rotation.md
tbrezot, line 1: Great design documentat!

Auto-rotation is an extremely complex topic (as in "it touches everything and thus impacts everything"). After reading this document I am left wondering with a few things:
- it seems useful to give the set of all rotations of a given key an *identity* (in Google's terms, this is a key-set, we could also call it key-chain to be more explicit about its structure; I will be using key-set in what follows), since from the user point of view, all the keys generated by auto-rotation are actually the same key. In a first approximation, it seems the user should not have to know more than that, which brings up the key selection issue.
- It seems needed to be able to: 
  - encrypt with a key-set (which uses its latest key)
  - encrypt with a specific key. We have the IDs, but since they are automatically and randomly generated, I don't think they can be of much use. We could instead have a special syntax to describe the version of the key to use in a key-set (e.g. `key-set-name@version`)
  - decrypt with a key-set (which tries each key in the key-set until it can successful decrypt)
  - decrypt with a specific key (same as encrypt)
  - decrypt with a key-set with a temporal hint (for example if we know the creation date of a ciphertext there is no need to attempt decrypting it with newer keys in the key-set).

  In Covercrypt we internally use similar techniques for key-rotation: each right is associated to a key-set which is implemented as a singly-linked list (why do you need a doubly-linked list in the KMS? It seems we never want to traverse it in chronological order but always in anti-chronological order when attempting to decrypt a document). We therefore have a key-set ID (the right), and each key in the set can be selected by its distance to the latest (the depth of the smallest prefix of the chain containing it).

- it is never made explicit here but it seems natural to prevent rotating keys that are not activated or expired.

tbrezot (commented):
I could not go through all diffs, but thanks to your documentation I could grasp enough of the idea to have a few questions (cf my comment of the doc). I will complete the review once we have discussed those points.

crate/interfaces/src/stores/objects_store.rs
Manuthor: https://github.com/Cosmian/kms/issues/983

crate/kmip/src/kmip_1_4/kmip_operations.rs
Manuthor: The `ReCertify` struct has been corrected to make all fields `Option<...>` per the spec. All fields in Table 188 have REQUIRED = "No". The `From<kmip_2_1::ReCertify> for ReCertify` conversion now directly maps `Option` to `Option` without forced defaults — `None` stays `None`. No `PEM` defaulting occurs during type conversion; validation of the conditional requirement (type must be present if request is present) belongs in the operation handler.

crate/kmip/src/kmip_1_4/kmip_operations.rs
Manuthor: The KMIP 1.4 spec ([§4.8, Table 188](https://docs.oasis-open.org/kmip/spec/v1.4/kmip-spec-v1.4.html#_Toc490660848) — Re-certify Request Payload) says *"If omitted, then the ID Placeholder value is used by the server as the Unique Identifier."* Our server does not implement the ID Placeholder mechanism (a batch-processing feature).

**Fix:** The `ReCertify` struct now uses `Option<String>` for `unique_identifier`, correctly modeling the spec's REQUIRED = "No" semantics. `None` stays `None` — no empty-string hack. The operation handler (`crate/server/src/core/operations/recertify.rs` L92–97) explicitly rejects requests with a missing identifier:
```rust
let uid = request
    .unique_identifier
    .as_ref()
    .ok_or_else(|| {
        KmsError::InvalidRequest(
            "ReCertify: unique_identifier of the certificate to rotate is required".to_owned(),
        )
    })?;
```

crate/kmip/src/kmip_2_1/kmip_operations.rs
Manuthor: The doc-comment on that field (`"If omitted, the ID Placeholder value is used"`) was copied verbatim from the KMIP spec but is **wrong** for this implementation: our server does not implement the ID Placeholder mechanism and explicitly rejects requests with a missing `unique_identifier` (see the handler in `recertify.rs`). 

About ID Placeholder, server is not supporting them at all.

On the serde point: `Option<T>` already deserializes a missing JSON/TTLV field as `None` without `#[serde(default)]`, so the attribute is technically redundant. What do you suggest?

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: I'm not sure to get it. This attribute does exist https://docs.oasis-open.org/kmip/kmip-spec/v2.1/kmip-spec-v2.1.html#_Toc57115612

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: Indeed 👍

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: Well this attribute is also defined in KMIP https://docs.oasis-open.org/kmip/kmip-spec/v2.1/kmip-spec-v2.1.html#_Toc57115608

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: Correct, thanks. Changing to i64 (du to another comment)

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: Good catch thank you. HSM keys do not have KMIP attributes, we should use PKCS11 flags instead. Probably in another issue

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: Indeed, will raise an error for now

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: IMO, the attribute is None for the first key and we can detect the first key of the key-set with the absence of this attribute

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: The latest key has no ReplacementObjectLink and this absence signals the chain origin

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: Removed.

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: Fixed. Same for the user decryption keys

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: A certificate is generated but without refreshing the key pair

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: It is the same behavior as wrapped key renewal

documentation/docs/kmip_support/key_auto_rotation.md
Manuthor: > It seems useful to give the set of all rotations a key-set identity.

Agree. Our current design uses the `x-rotate-name` vendor attribute as the key-set identity. All rotations of a key share the same `x-rotate-name` value, which serves as the stable identifier the user interacts with. 

Let me paraphrase just to be sure I've understood:

- **Encrypt with key-set** → `Locate` by `x-rotate-name` + `x-rotate-latest=true` → returns the current active key → `Encrypt`
- **Encrypt with specific version** → Use the key's UID directly (or `x-rotate-name` + `x-rotate-generation=N`)
- **Decrypt with key-set** → Client-side: try latest first, walk `ReplacedObjectLink` chain backward (or server could try each key — not yet implemented)
- **Decrypt with temporal hint** → Use creation-date metadata to narrow the candidate set

The `key-set-name@version` syntax is a great suggestion. We could support it as a UID alias in the `Locate` / operation path.

> It seems we never want to traverse it in chronological order but always in anti-chronological order when attempting to decrypt.

**Answer:** The doubly-linked list (both `ReplacedObjectLink` and `ReplacementObjectLink`) is mandated by [KMIP 2.1 §6.1.46 (Re-key)](https://docs.oasis-open.org/kmip/kmip-spec/v2.1/os/kmip-spec-v2.1-os.html#_Toc57115678) and [§11.28 (Link Type Enumeration)](https://docs.oasis-open.org/kmip/kmip-spec/v2.1/os/kmip-spec-v2.1-os.html#_Toc57115808), Table 464:
- `Replacement Object Link` (value `00000106`) on the old key (points forward to the new key)
- `Replaced Object Link` (value `00000107`) on the new key (points backward to the old key)

Use cases for forward traversal:
1. **Admin cleanup**: "find all successors of this key and revoke/destroy them" (e.g., security incident response)
2. **Audit trail**: walk the chain forward to show the full rotation history
3. **Migration**: when upgrading a key to a new algorithm, walk forward to find the latest version

In practice, decryption indeed only needs backward traversal (newest → oldest). But the KMIP spec requires both directions, and both are cheap to maintain (one link per key).

> It seems natural to prevent rotating keys that are not activated or expired.

**Answer:** Agreed and partially implemented:
- ✅ Deactivated keys → error
- ✅ Destroyed keys → error (can't even retrieve)
- ⚠️ PreActive keys → currently allowed (debatable — a PreActive key hasn't been used yet, so rotation might be premature)
- ⚠️ Compromised keys → currently allowed (might want to prevent rotation of compromised keys to avoid confusion)

crate/server/src/core/wrapping/wrap.rs
Manuthor: Correct, it will be fixed

crate/server/src/core/kms/kmip.rs
Manuthor: Will be fixed

documentation/docs/kmip_support/key_auto_rotation.md
tbrezot: Thanks for your answer. Your rephrase is correct. Thanks for pointing out the KMIP requirements (since otherwise the use-cases you mention seem to work by walking the chain backward from the latest key of the key-set that should always be "locatable" given the `owm` of a given key in the key-set).

About forbidding rotation of non-active keys, I was talking about *auto-rotation*. But even then it is unclear what is the sensible thing to do and would require more discussion imho.

Related to that is the interaction of rotation with the key-set view of the system: what happens if the user rotates an older version of a key? Does it fork the chain? Does it truncate it? Imho rotation should always target key-set (or equivalently, manual rotation of keys that are not marked as _latest_ should not be authorized)

crate/interfaces/src/stores/objects_store.rs
Manuthor: Good suggestion. Done in efc4239db695299cb1abef376cc0ff656f52aa8f where the wrapping_key_id column is added. All back-end should be able to implement it (without JSON-based object)

crate/server_database/src/stores/redis/redis_with_findex.rs
Manuthor: Has been rewritten by indexing wrapped key in a4ec0aeeabed3de280b28826dada3b8f832bf2a7

crate/server/src/config/command_line/clap_config.rs
```diff
@@ -213,6 +215,17 @@ pub struct ClapConfig {
...
+    /// Interval in seconds between background auto-rotation checks.
+    /// Set to 0 (default) to disable the auto-rotation background task.
+    #[clap(long, default_value = "0", verbatim_doc_comment)]
+    pub auto_rotation_check_interval_secs: u64,
+
+    /// Maximum number of keys the server will try when decrypting with a keyset
+    /// identifier (walking the rotation chain from newest to oldest).
+    /// Prevents unbounded chain traversal.
+    #[clap(long, default_value = "100", verbatim_doc_comment)]
+    pub keyset_decrypt_max_attempts: u32,
```
tbrezot: Since the number of keys hold by the KMS is finite, unbounded chain traversal may only be problematic when encountering a cycle. I propose to implement cycle-detection instead or limiting the depth of the traversal it could prevent decrypting very old ciphertexts (or ciphertexts which associated key-set is often rotated.

In term of performance, long chains still may be an issue, especially if we are talking about Covercrypt (note that the latest Covercrypt user secret keys may always be able to decrypt old ciphertexts since they hold the whole history, depending on how you refreshed them (there is a boolean to indicate whether or not you wish to preserve the history)) or KEM encapsulations. Rather than preventing decryption, I would authorize unbounded traversal and return to the user both the decrypted data and the depth at which the successful decryption happened (potentially issuing a warning client-side if this depth is greater than a configurable threshold).

crate/server/src/config/command_line/clap_config.rs
```diff
@@ -213,6 +215,17 @@ pub struct ClapConfig {
     #[clap(flatten)]
     #[serde(rename = "kmip")]
     pub kmip_policy: KmipPolicyConfig,
+
+    /// Interval in seconds between background auto-rotation checks.
+    /// Set to 0 (default) to disable the auto-rotation background task.
+    #[clap(long, default_value = "0", verbatim_doc_comment)]
+    pub auto_rotation_check_interval_secs: u64,
```
tbrezot, line 225: We may want to prevent the user from configuring small values as it would create quite the churn KMS-side.

crate/server/src/core/cover_crypt/rekey_keys.rs
tbrezot, line 1: What a nice sight to see this omnipresent `privileged_users` disappear 👍

crate/server/src/core/operations/attributes/modify.rs
```diff
@@ -49,10 +49,20 @@ pub(crate) async fn modify_attribute(
...
+                    .to_owned(),
+            ));
+        }
+        Attribute::RotateName(name) if name.contains('@') => {
+            return Err(KmsError::InvalidRequest(
+                "ModifyAttribute: rotate_name must not contain '@' (reserved for keyset versioning)"
+                    .to_owned(),
             ));
         }
         _ => {}
```
tbrezot, line 69: This pattern is repeated quite a lot, I think it would gain to factorize it somewhere (even though I am not entirely sure of what is the best way to do it).

After reviewing the KMIP specification the past few weeks, I think they have a point: in big systems with lots of moving parts, it is sometimes easier to replace hard-coded logic by metadata structures holding such logic. For example, instead of checking at specific places whether the user is authorized to modify a given attribute (which implies it is easy *not* to check it), attribute-modification functions could internalize and systematize those checks by leveraging metadata structures that can be queried to retrieve information about attributes types. For example here we could have the metadata in the form of a set containing (attribute types, property) pairs, and have the modification function do something like:
```rust
fn modify_attr(kms: &KMS, type: &str, attr: Attribute, ...) -> KResult<()> {
    if !kms.attr_properties.is_member((type, AttributeProperty::UserModifiable)) then
        return KError(...);
    }

    // This part is harder to design, it could be an ad-hoc method. 
    if kms.attr_properties.is_valid_name(type, attr.name) then {
        return KError(...);
    }

    // perform the modification
    ...
} 
```
The same goes for many properties about attributes and keys that are checked at many different places. Adding a level of reflexivity in the KMS could allow for significant systematization and factorization.

Of course this is not an issue to fix right now, but rather some comment that may nurture future discussions.

crate/server/src/core/operations/attributes/set.rs
```diff
@@ -140,6 +191,40 @@ pub(crate) async fn set_attribute(
...
+        let label = format!("{rotate_name}::0::{key_id}::latest");
+        trace!(
+            "SetAttribute: writing CKA_LABEL '{}' on HSM key '{}'",
+            label,
+            owm.id()
+        );
+        kms.database.set_key_label(owm.id(), &label).await?;
+    } else if let Some(interval_secs) = hsm_rotate_interval_secs {
+        let today = OffsetDateTime::now_utc().date();
+        let days = interval_secs / 86400;
```
tbrezot: Where does this number come from? Maybe write it as `(24 * 3600)` since those are known values.

crate/server/src/core/operations/attributes/set.rs
```diff
@@ -140,6 +191,40 @@ pub(crate) async fn set_attribute(
...
+        let end_date = today + time::Duration::days(days);
```
tbrezot: Is the KMS-clock granularity days? Or is there a missing second remainder?

crate/server/src/core/operations/key_ops/mod.rs
```diff
@@ -15,21 +15,64 @@ use cosmian_kms_server_database::{
...
+        if !has_permission
+            && !users.iter().any(|u| u == user)
+            && user != kms.params.default_username
+        {
+            kms_bail!(KmsError::Unauthorized(
+                "User does not have create access-right.".to_owned()
+            ))
+        }
+    }
+    Ok(())
```
tbrezot: This `OK` default really tripped me up, consider adding a comment to help the reader:
```suggestion
    // If no privileged user was set, all users have the `Create` right.
    Ok(())
```

crate/server/src/core/operations/key_ops/mod.rs
```diff
@@ -15,21 +15,64 @@ use cosmian_kms_server_database::{
...
+        let has_permission = user_has_permission(user, None, &KmipOperation::Create, kms).await?;
+
+        if !has_permission
+            && !users.iter().any(|u| u == user)
+            && user != kms.params.default_username
+        {
+            kms_bail!(KmsError::Unauthorized(
+                "User does not have create access-right.".to_owned()
+            ))
+        }
```
tbrezot: Composing negation of boolean values is always risky, consider rewriting it the other way around (also you can save a DB round-trip in case the user is privileged or is the default user -- this really is a nitpick as I doubt this will make any difference in practice) :

```suggestion
        if user == kms.params.default_username
             || users.iter().any(|u| u == user)
             || user_has_permission(user, None, &KmipOperation::Create, kms).await?
        {
            Ok(())
        } else {
            kms_bail!(KmsError::Unauthorized(
                "User does not have create access-right.".to_owned()
            ))
        }
```

crate/server/src/core/operations/rekey/common.rs
```diff
@@ -0,0 +1,889 @@
...
+    }
+    Ok(())
+}
+
+/// Retrieve all eligible objects matching the given identifier, filtered by state and type.
+///
+/// Filters by:
+/// - State: `Active` or `Deactivated` (per KMIP §6.1.46 — `Wrong_Key_Lifecycle_State` is NOT
+///   listed in the Re-Key error table, meaning Deactivated keys are eligible for rotation).
+///   `PreActive`, `Compromised`, `Destroyed`, and `Destroyed_Compromised` keys are rejected.
```
tbrezot: What is the rational behind forbidding the rotation of compromised keys? The way I see it, if a key is compromised, you want to rotate it as soon as possible! (on the other hand, forbidding rekeying pre-active and destroyed objects makes sense)

crate/server/src/core/operations/rekey/common.rs
```diff
@@ -0,0 +1,889 @@
...
+        replacements: &[ReplacementObject],
+    ) -> impl std::future::Future<Output = KResult<()>> {
+        default_finalize_dependants(kms, user, candidates, replacements)
+    }
+
+    /// Step 8: Build the KMIP response from the completed replacements.
+    fn build_response(&self, replacements: &[ReplacementObject]) -> Self::Response;
+}
+
+/// Default implementation for [`RekeyOperation::persist_new_key`].
```
tbrezot: Then write it as a default instead (i.e. in the trait definition).

crate/server/src/core/operations/rekey/common.rs
tbrezot, line 1: Would it be possible to perform all DB modifications in an atomic fashion? It seems the current trait splits them into several atomic operations.

Also, I don't recall the KMS DB trait to expose a guard operations allowing to fail a transaction conditionally to the concurrent modification of an object. What happens if a manual rotation is performed concurrently with an automatic rotation? What about two concurrent manual rotations requests? We may want to land this MR as soon as possible, but I am not convinced of the concurrency model. This feels like it may require profoundly refactoring the current DB abstraction.

crate/server/src/core/operations/rekey/keypair/sql.rs
```diff
@@ -0,0 +1,507 @@
...
+        )?;
+
+        enforce_privileged_user(kms, user).await?;
+
+        let uid_or_tags = request
+            .private_key_unique_identifier
+            .as_ref()
+            .ok_or(KmsError::UnsupportedPlaceholder)?
+            .as_str()
+            .context("ReKeyKeyPair: the private key unique identifier must be a string")?;
```
tbrezot, line 92: It seems that properly parsing the tags once upon receiving a request into an algebraic data-type like:
```
enum Handle {
    Tag(Vec<String>),
    Hsm(KeyID),
    Owm(ObjectID),
}
```
would simplify a lot of code.

crate/server/src/core/operations/rekey/keypair.rs
```diff
@@ -0,0 +1,507 @@
...
+            // Validate no crypto param changes
+            validate_no_crypto_param_change(
+                owm.attributes(),
+                [
+                    request.common_attributes.as_ref(),
+                    request.private_key_attributes.as_ref(),
+                    request.public_key_attributes.as_ref(),
+                ],
+                "ReKeyKeyPair",
+            )?;
```
tbrezot: Why not reusing the ones of the object and ignoring those of the request directly?

crate/server/src/core/operations/rekey/keypair.rs
```diff
@@ -0,0 +1,507 @@
...
+        if uid_or_tags.starts_with("hsm::") {
+            return Err(KmsError::NotSupported(
+                "Re-Key Key Pair is not supported for HSM-managed keys. \
+                 Use PKCS#11 vendor tools or the HSM administration console \
+                 to manage HSM key lifecycle."
+                    .to_owned(),
+            ));
+        }
+
+        for owm in retrieve_eligible_keys(kms, uid_or_tags, ObjectType::PrivateKey).await? {
```
tbrezot: It seems this loop silently ignores all matches after the first valid one.

I raised a similar issue of key selection for operations like `encrypt` and `sign`. Key selection should be treated uniformly.

Note that the current key selection procedure prevents using a compromised key for decryption or verification, which is not the behavior specified in KMIP 2.1. To correct this, remove the check from `crypto_op.rs:248` and add a state check to the implementation of `CryptoOpSpec` accordingly to the operation.

crate/server/src/core/operations/rekey/keypair.rs
```diff
@@ -0,0 +1,507 @@
...
+                    uid: old_sk_uid,
+                    object_type: ObjectType::PrivateKey,
+                    owm,
+                },
+                RotationCandidate {
+                    uid: old_pk_uid,
+                    object_type: ObjectType::PublicKey,
+                    owm: old_pk_owm,
+                },
+            ]);
```
tbrezot: Relying on a vector for variadicity is rather brittle. Consider adding an associated type to describe the nature of the expected candidates and replacements. For example for a keypair:
```
    type Candidate = (RotationCandidate, RotationCandidate);
    type Replacement = (ReplacementObject, ReplacementObject);
```

tbrezot (commented):
A lot of effort has been put into this MR, and it greatly improves the initial implementation. I still have a few comments though:
- a few nitpicks that are trivial to correct;
- maybe an issue concerning the key-selection procedure;
- some deeper concerns about concurrency management, which are orthogonal to this feature and may be addressed in a dedicated MR.
 
Also, since it is structuring, it might gain from requesting a review from someone else. In any case, I may spend more time on it if you delay the merge (since it is huge, I could not read everything).

crate/server/src/config/command_line/clap_config.rs
Manuthor: Fix in 621e4550ee301fd8fc6a901c7765d639cb2b9cc6

crate/server/src/config/command_line/clap_config.rs
Manuthor: Fix in c4435ba2e3301511be1cf2b63414a394b0a3dab8

crate/server/src/core/operations/attributes/modify.rs
Manuthor: Referenced in #1004

crate/server/src/core/operations/attributes/set.rs
Manuthor: Fix in a76d76056

crate/server/src/core/operations/attributes/set.rs
Manuthor: There is indeed an issue about Rotate Interval (defined as seconds in KMIP and indirectly defined in days for PKCS#11):
- KMIP 2.1 - Rotate Interval unit (seconds):
  * [§4.51](http://docs.oasis-open.org/kmip/kmip-spec/v2.1/os/kmip-spec-v2.1-os.html#_Toc57115610) Rotate Interval (attribute definition): 
  * [§11.23](https://docs.oasis-open.org/kmip/kmip-spec/v2.1/os/kmip-spec-v2.1-os.html#_Toc57115803) Interval (normative type definition - "Four-byte long (32 bit) unsigned numbers in seconds")
- PKCS#11 v3.0 does not have a Rotate Interval. The relevant section for date-granularity (day, not seconds) is:
  * [3.4](https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html#_Toc29976544): CK_DATE ({ year[4], month[2], day[2] } — day granularity only, used by CKA_START_DATE / CKA_END_DATE)
  * [4.7.2](https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html#_Toc29976579): Common Key Attribute Table 23: CKA_START_DATE, CKA_END_DATE both typed CK_DATE

Fix HSM RotateInterval scheduling bugs: reject sub-day intervals in 17c68bf58

crate/server/src/core/operations/key_ops/mod.rs
Manuthor: Done in a573d6a182c6f52320fe3978345037e1249f5854

crate/server/src/core/operations/rekey/common.rs
Manuthor: Done in c3c240d9d

crate/server/src/core/operations/rekey/common.rs
Manuthor: Done in d1ff3d960

crate/server/src/core/operations/rekey/common.rs
Manuthor: Referenced in #1006

crate/server/src/core/operations/rekey/keypair/sql.rs
Manuthor: Referenced in #1007

crate/server/src/core/operations/rekey/keypair.rs
Manuthor: You're right - and that's already what happens (generate_replacement):
```
let common_attrs = build_generation_attributes(sk_candidate.owm.attributes(), ...);
```

The validate_no_crypto_param_change is defensive

crate/server/src/core/operations/rekey/keypair.rs
Manuthor: Done in 9b3a0e39a

crate/server/src/core/operations/rekey/keypair.rs
Manuthor: Done in c046e391a

crate/server/src/core/operations/rekey/symmetric/hsm.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// Returns the stable base UID (no generation suffix).
+    fn full_base_uid(&self) -> String {
+        format!("{}::{}::{}", self.prefix, self.slot_id, self.base_id)
+    }
+}
+
+/// Parse an HSM key UID into its components.
+///
+/// Expected format: `{prefix}::{slot}::{base_id}` or `{prefix}::{slot}::{base_id}@{gen}`.
+fn parse_hsm_uid(uid: &str) -> KResult<ParsedHsmUid> {
```
grydz, line 43: If there is a strict definition of HSM UID, do you think it would be better to use a regular expression with groups for parsing?

crate/server/src/core/operations/rekey/symmetric/hsm.rs
```diff
@@ -0,0 +1,336 @@
...
+        // Fall back gracefully if CKA_LABEL metadata is absent.
+        let (rotate_name, old_rotate_gen) = (
+            old_attrs.rotate_name.clone(),
+            old_attrs.rotate_generation.unwrap_or(old_gen),
+        );
+
+        // Read rotation interval in days from the old key's attributes.
+        // For HSM keys, rotate_interval is not stored in PKCS#11 as a KMIP attribute
+        // (HsmStore::update_object is a no-op); instead it is reconstructed at
+        // retrieve-time from CKA_START_DATE / CKA_END_DATE as (end − start) × 86400 s.
```
grydz:
```suggestion
        // retrieve-time from CKA_START_DATE / CKA_END_DATE as (end − start) × (24 * 60 * 60) s.
```

crate/server/src/core/operations/rekey/symmetric/hsm.rs
Manuthor: There is an issue about using a regex since we have currently 2 formats for HSM keys:
- Old format: hsm::<slot_id>::<key_id>[@gen] → prefix = "hsm"
- New format: hsm::model::<slot_id>::<key_id>[@gen] → prefix = ["hsm::<model>"]
IMHO, we should remove the old format in a future release, so I will add a TODO in the code to fix it properly later

PR #1005, opened 2026-06-15
feat: add ai skills

### LINKED ISSUE #986: Split AGENTS.md, add skills, optimize token usage
tl;dr : the KMS codebase is huge, AI is becoming more and more useful to work, but also more and more expensive. Even if we forget about the financial cost of LLMs, an AI context file that's **always** loaded should be around 500 lines. The KMS's is 4000.

Some ideas : 
- Create KMS Skills
- Split instructions to file scoped files (exp: UI instructions should be discarded when the LLM is reading rust files)
- https://github.com/safishamsi/graphify graphify is becoming the reference in token usage optimization and a lot of other related goals

Manuthor:
- Create KMS skills for copilot-cli and VSCode Chat.
- Create minimal required Copilot-instructions

Closes #986


.github/copilot-instructions.md
HatemMn, line 1: Might be a mistake, but this file should better be omitted, Agents md is well taken into account

.github/skills/docs-writer/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+4. **User-provided**
+
+**If none applies**: Leave `"<result>"` placeholder and report: "Example not sourced — could not find test assertion, live KMS output, derivation, or user value for: [description]". **Never invent examples.**
+
+## Contextual Awareness
+
+- Use existing pages as context for tone, style, and terminology.
+- Do NOT copy content from them unless explicitly asked.
+- Do not consult external websites unless the user provides a link.
+- For algorithm OIDs and spec references: use rfc-editor.org (IETF), OASIS pages (KMIP), csrc.nist.gov (FIPS), oidref.com (X.509/ASN.1).
```
HatemMn, line 99: Aren't those contradictory ? Also, isn't not consulting the web a little bottleneck ? Usually our current agents don't tend to search the internet unless asked to - but not fetching the web is systematically equal to relying on training data

.github/skills/adr/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+If any required input is missing, ask before proceeding.
+
+## ADR Template
+
+```markdown
+---
+title: "ADR-NNNN: [Decision Title]"
+status: "Proposed"
+date: "YYYY-MM-DD"
+authors: "[Stakeholder Names/Roles]"
```
HatemMn, line 41: do KMS Stakeholders have access to coding tools ?

.github/skills/fips-audit/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+| Key size enforcement | ✅ / ⚠️ N issues |
+| OpenSSL provider init | ✅ / ⚠️ N issues |
+| Entropy / RNG | ✅ / ⚠️ N issues |
+
+### Findings
+[Severity + description + file + line + fix for each issue]
+
+### Clean Areas
+[List areas with no issues]
+```
```
HatemMn, line 141: Honestly I think this [List areas with no issues]  2 line will increase token consumption (output) with no user benefit

.github/skills/fips-audit/SKILL.md
```diff
@@ -0,0 +1,146 @@
+---
+name: fips-audit
+description: Review Rust crypto code for FIPS 140-3 compliance: approved algorithms, key sizes, non-fips gating discipline, and OpenSSL provider init. Use when touching crate/crypto/ or algorithm selection.
```
HatemMn: shouldn't the number 140-3 be mentionned in the file name to avoid conflicts of an absolete skill when adding newer versions some day ?

p0wline (approved):

.github/prompts/adr.prompt.md
HatemMn, line 1: Pretty much a transversal comments: which of these skills were actually used in an actual project and which are experiments ? Those who derive from the old AGENTS.md can be considered viable. Can't talk about the others. Can we maintain some internal doc page (shall it be a slack channel) to take notes about our thoughts/experiences since all of this is still experimental ?

.github/skills/adr/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+## ADR Storage
+
+Save ADRs at: `documentation/docs/adr/adr-NNNN-[title-slug].md`
+
+Where `NNNN` is the next sequential 4-digit number. Check existing files to determine the next number:
+
+```bash
+ls documentation/docs/adr/ 2>/dev/null | sort | tail -5
+```
+
```
HatemMn, line 19: This set of lines means that this should be often done in the codebase, committed and pushed. Is this a new work procedure ? Maybe it should be better explained and documented at CONTRIBUTING.md level and even at README level ? I didn't have much time to read about ADR but it looks promising. Commoners tend to do a PRD but I am highly against that approach and this one seems like the correct alternative

.github/skills/ci-efficiency/SKILL.md
HatemMn, line 1: IMO, most CI time is wasted because of build times, un-optimized stuff and especially redundant testing, while the two first ones are huge problems shoudn't eliminating redundant tests should be the core search of this skill ?

Redundant is not **always** repetition: it can be two tests that use test_vectors but end up doing same or similar flows under different names (I already deleted one of them because of this reason)

.github/skills/ci-fix/SKILL.md
HatemMn, line 1: 1/4 on my failed CI runs are because of NIX hashes, would it be intresting to add the command 

```sh
 .github/scripts/release/update_hashes.sh [failed job link]
```

to list of cummon fails along with clippy and cargo fmt ?

Also, maybe it's better to give him this command : 

```sh
 cargo clippy --workspace --all-targets --all-features -- -D warnings && cargo fmt --all
```

Another last suggestion if you agree with me: telling the agent, when a test fails, to run only the failed test after fixing whatever he did to check for sanity. I don't know why, mine runs all 144 (or 200+) tests just after editing one file, despite that file not affecting anything else

```
for sanity checks, only run the tests that directly exercise the changed code. When code changes affect only a certain scope, target that scope. Don't run the full test suite unless this is considered to be relevant according to the changes you made.
```

.github/skills/conventional-commit/SKILL.md
HatemMn, line 1: Recent new agents, know this information very well because it's heavely present on the training data. Do we really need a full skill ?

Image

.github/skills/kms-test-vector/SKILL.md
HatemMn, line 1: ⚠️ **important general comment**
This is one "task" that I myself used to never even do and it was an amazing approach to have the default agent MD do it at ALL projects if it's relevant

This question apply to a lot of other skills, since we now split the AGENT MD file into mutiple human actionnable skills, won't we lose the "superpower" of the coding agents ? Them knowing and acing the codebase by default was the whole goal of having that huge file, I remember researching ways that make agents auto direct themselves to sub-instruction files, but I didn't study this yet

.github/skills/react-ant-patterns/SKILL.md
```diff
@@ -0,0 +1,277 @@
+---
+name: react-ant-patterns
+description: React 19 + Ant Design 5 + Tailwind 4 + Vite 7 patterns for the KMS Web UI: WASM integration, FIPS guards, data-testid placement, Playwright companion. Use as a reference for UI development patterns.
+---
+
+# React + Ant Design Patterns for the KMS UI
+
+Reference for KMS UI-specific patterns. Stack: **React 19 + Vite 7 + Ant Design 5 + Tailwind CSS 4 + pnpm**.
+
```
HatemMn, line 9: AntD are very up-to-date twoard agentic coding, so I think it's more useful to leverage that,
1) first they support llm.txt (https://ant-design.antgroup.com/docs/react/llms) so we can give tell the agent something like this : 
```
Fetch [https://ant.design/llms.txt](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and **[actual instructions]**
```
2) AntD Mcp server : https://ant-design.antgroup.com/docs/react/mcp (didn't try it yet)

**General comment :**

A "do it all" solution for outdated libraries (very cummon problem with javascript world since agents love to pull their outdated training data) is this MCP server : **[io.github.upstash/context7](https://context7.com/upstash/context7)**
It's top rated and pulls new docs from any language's library and is heavely used among devs

It's downloadable in vscode btw

CONTRIBUTING.md
```diff
@@ -126,6 +126,30 @@ writing), create a symbolic link so it picks up the agent instructions automatic
...
+| **Commit scope** | Minimal, focused changes — don't refactor unrelated code alongside a bug fix |
+
+For detailed Rust design patterns and idiomatic conventions specific to this codebase, see
+[`.github/prompts/rust-patterns.prompt.md`](.github/prompts/rust-patterns.prompt.md).
+For FIPS gating rules and the algorithm allow-list, see
+[`.github/prompts/fips-audit.prompt.md`](.github/prompts/fips-audit.prompt.md).
+For React/TypeScript/WASM conventions, see
+[`.github/prompts/react-ant-patterns.prompt.md`](.github/prompts/react-ant-patterns.prompt.md).
+
+---
```
HatemMn, line 151: while I agree with all what's cited in here, will human contributors really this machine sounding phrasing ? This file is meant for human cotributors, maybe we can make the table more light ?

.github/instructions/docs.instructions.md
```diff
@@ -0,0 +1,51 @@
+---
+name: 'Documentation'
+description: 'MkDocs documentation conventions (Diátaxis framework)'
+applyTo: '{documentation,cli_documentation}/**/*.md'
+---
+
+# Documentation rules
+
+## Framework: Diátaxis
```
HatemMn, line 9: spelling mistake ?

HatemMn (changes requested):
Very good refactor ! some "general" questions still rise and should be discussed : 

1. A big number of skills would often discourage people to even browser them, do we really need all of these ?
2. Shrinking the AGENTS.md strips intellingence from the agent instead of just asking him to use the correct tool (read below, a comment talks about this in details)
Also, diffs tell **+4,826-311** , so actually almost ALL the new files are 'new' and not content redirection.
I suggest raising an issue that marks this PR and asks to investigate how the split was made, or research it here if you think this is useful ?
3. I honestly didn't understand the nuance between instructions and skills
4. No MCP server usage is documented. This isn't a mistake nor a problem per-se. But We can take it into account for later on

.github/copilot-instructions.md
Manuthor: `copilot-instructions.md` and `AGENTS.md` serve **different purposes** per GitHub/VS Code docs:

- `.github/copilot-instructions.md` → injected into every **Copilot Chat** request (IDE panel, inline suggestions). See [Adding repository custom instructions](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions).
- `AGENTS.md` → read by **autonomous agents** (Copilot agent mode, Claude, Codex) for task execution. See [Use an AGENTS.md file](https://code.visualstudio.com/docs/agent-customization/custom-instructions#_use-an-agentsmd-file).

Both are needed. The `copilot-instructions.md` is intentionally a concise summary (~2 pages) while AGENTS.md is the full reference. Removing `copilot-instructions.md` would make Copilot Chat lose project context in non-agent mode.

.github/skills/docs-writer/SKILL.md
Manuthor: The rule "Do not consult external websites unless the user provides a link" prevents the agent from **hallucinating URLs** or fetching unreliable sources. This is a known best practice from the [VS Code tips for writing effective instructions](https://code.visualstudio.com/docs/agent-customization/custom-instructions#_tips-for-writing-effective-instructions) — agents should ground answers in local context (codebase, tests, specs in `kmip/`).

However, point taken that it could be softened. I've kept it because the spec references (IETF, OASIS, NIST) are listed explicitly as allowed sources in the same file. The "contradiction" is intentional: it means "don't browse random sites, but DO use the authoritative sources listed below."

.github/skills/adr/SKILL.md
Manuthor: ADRs are written by **developers** and committed to the repo. Stakeholders are listed as "roles affected by the decision" (e.g. "operators", "security auditors"), not as people who write or review the ADR in the IDE. The purpose is documenting *who is impacted*, not requiring those people to use coding tools.

This follows the standard ADR practice from [Michael Nygard's original ADR proposal](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions).

.github/skills/fips-audit/SKILL.md
Manuthor: Agreed — removed. The "Clean Areas" section added output tokens with no actionable value for the developer.

.github/skills/fips-audit/SKILL.md
Manuthor: Skill file names in VS Code Copilot follow the convention of short, descriptive identifiers matching the directory name (see [Agent Skills docs](https://code.visualstudio.com/docs/agent-customization/agent-skills#_create-a-skill)). The FIPS version is specified inside the skill content, not the filename.

If FIPS 140-4 arrives, we update the skill content (algorithm list, provider init, etc.) — the skill's *purpose* (auditing crypto compliance) won't change, only the spec version it checks against. A rename would break all references in AGENTS.md and `.github/prompts/`.

.github/prompts/adr.prompt.md
Manuthor: Good question. The architecture is:

1. **Skills derived from AGENTS.md** (proven in practice): `ci-fix`, `kms-sync-rules`, `kms-test-vector`, `kms-changelog`, `security-review`, `fips-audit`, `kmip-compliance`, `rust-patterns`
2. **New skills** (added proactively based on team needs): `adr`, `docs-writer`, `ci-efficiency`, `pre-release`, `react-ant-patterns`, `playwright-kms`

For tracking usage/feedback, a Slack channel or a `documentation/docs/adr/` page on "AI Skills adoption" would work. Happy to create an issue for this.

.github/skills/adr/SKILL.md
Manuthor: Agreed that ADR as a practice should be documented. It's not a "new tool dependency" — it's a lightweight way to record *why* architectural decisions were made (instead of losing that context in Slack threads).

I'll add a brief mention in CONTRIBUTING.md under a "Decision Records" section. The ADR skill is the agent's tool for creating them; the human process is simply: "when making a significant architectural choice, run `/adr` or manually create a file in `documentation/docs/adr/`."

.github/skills/ci-efficiency/SKILL.md
Manuthor: Good point — added "Redundant test detection" as item #7 in the ci-efficiency skill. It now explicitly calls out semantic overlap (not just textual repetition) and test-vector tests exercising similar flows under different names.

.github/skills/ci-fix/SKILL.md
Manuthor: All three suggestions applied:

1. **NIX hashes** — added `.github/scripts/release/update_hashes.sh` as the recommended fix for hash mismatches
2. **Combined command** — updated the local verify step to `cargo clippy --workspace --all-targets --all-features -- -D warnings && cargo fmt --all`
3. **Targeted test runs** — added explicit guidance: "only run the tests that directly exercise the changed code. When changes affect only a certain scope, target that scope. Do not run the full test suite unless the change is cross-cutting."

.github/skills/conventional-commit/SKILL.md
Manuthor: Fair point — models do know conventional commits well from training data. The skill adds value in two ways:

1. **Repo-specific scopes**: Our allowed scopes (`kmip`, `crypto`, `server`, `cli`, `ui`, `ci`, `nix`, `docs`, `deps`) aren't universal — without the skill, agents guess wrong scopes.
2. **Consistency enforcement**: The skill enforces our specific `<type>(<scope>): <description>` format with the pre-commit hook validation in mind.

That said, it's one of the lighter skills and could be merged into the general instructions if we want to reduce skill count. Open to either approach.

.github/skills/kms-test-vector/SKILL.md
Manuthor: This is the key architectural concern and it's well-founded. The answer is that **splitting does NOT remove agent context** — it's layered:

1. **`AGENTS.md`** remains the full reference loaded by every agent session. It still contains the execution workflow (Section 5) that references skills by name. Agents see it automatically.
2. **`.github/instructions/*.instructions.md`** files are **auto-applied** based on `applyTo` glob patterns — agents get them injected whenever they touch matching files. See [Use .instructions.md files](https://code.visualstudio.com/docs/agent-customization/custom-instructions#_use-instructionsmd-files).
3. **Skills** (`.github/skills/*/SKILL.md`) are invoked on-demand via `/slash-commands`. They're for complex multi-step workflows that would bloat AGENTS.md if inlined. See [Agent Skills](https://code.visualstudio.com/docs/agent-customization/agent-skills).

So the "superpower" is preserved: AGENTS.md + auto-applied instructions give the agent full context. Skills add depth for specific tasks without inflating the always-loaded context window.

.github/skills/react-ant-patterns/SKILL.md
Manuthor: Great suggestions — applied. Added a new "AI-Assisted Development Resources" section to the react-ant-patterns skill with:

1. **AntD `llm.txt`**: `https://ant.design/llms.txt` — agents can fetch this for up-to-date component APIs
2. **AntD MCP server**: linked to https://ant-design.antgroup.com/docs/react/mcp
3. **Context7 MCP** (`io.github.upstash/context7`): generic library doc fetcher for when training data is stale

Re: MCP server configuration — agreed this should be documented separately (as noted in your general review comment #4). Will track as a follow-up.

CONTRIBUTING.md
Manuthor: The table content is identical to what's in AGENTS.md § "Cardinal rules" — it's intentionally shared between both audiences (human contributors + agents) so there's a single source of truth for coding standards.

The table format is concise (one rule per row) and scannable. Happy to soften the intro line from "non-negotiable" to something lighter if that reads better for humans, but the rules themselves should stay as-is since they match what the CI enforces.

.github/instructions/docs.instructions.md
Manuthor: Not a spelling mistake — **Diátaxis** is the proper name of the documentation framework (from Greek διάταξις = "arrangement"). See https://diataxis.fr/

The accent is intentional and matches the official branding.

Manuthor:
Addressing the general review points:

**1. Too many skills?** — Skills use progressive loading: only the `name` and `description` from YAML frontmatter are discovered initially; the full body is loaded only when relevant. See [How Copilot uses skills](https://code.visualstudio.com/docs/agent-customization/agent-skills#_how-copilot-uses-skills). Having many skills does NOT bloat context. Only `.github/instructions/*.instructions.md` files with `applyTo` patterns are auto-injected.

**2. Shrinking AGENTS.md?** — AGENTS.md was **not** shrunk — it's unchanged in content. The skills are *additive*. The +4826 lines are all new skill files that provide depth beyond what AGENTS.md covers. AGENTS.md remains the always-loaded context.

**3. Instructions vs Skills nuance:** The VS Code docs explicitly address this distinction in [Agent Skills vs custom instructions](https://code.visualstudio.com/docs/agent-customization/agent-skills#_agent-skills-vs-custom-instructions):
- **Instructions** (`.github/instructions/`) → define coding standards/guidelines, auto-applied via `applyTo` glob. See [Use .instructions.md files](https://code.visualstudio.com/docs/agent-customization/custom-instructions#_use-instructionsmd-files).
- **Skills** (`.github/skills/`) → teach specialized capabilities and workflows, loaded on-demand via `/slash-commands`. They can include scripts, examples, and resources alongside instructions.

**4. MCP servers?** — Agreed, not in scope here. Opened as follow-up topic. The AntD MCP reference was added to the react-ant-patterns skill.

---

**Applied fixes from this review:**
- Removed "Clean Areas" section from fips-audit (token waste)
- Added redundant test detection to ci-efficiency
- Added NIX hash script + targeted test guidance to ci-fix
- Added AntD llm.txt / MCP / context7 references to react-ant-patterns

.github/skills/fips-audit/SKILL.md
HatemMn: I think I was not clear enough, the rename is for **human** operator not AI. Anyone can miss out outdates, also _A rename would break all references in AGENTS.md and .github/prompts/._ and _The FIPS version is specified inside the skill content, not the filename._ , aren't those contradictory ?

Anyway I leave you to decide of this, not very important

.github/prompts/adr.prompt.md
HatemMn: Okay please also keep note of this comment and leave it open if possible :) that's the list I needed

Also side note, security-review / rust-patterns can be imporved using literatture as discussed in meetings

.github/skills/adr/SKILL.md
HatemMn: At the same time, is there a client demand to have those docs in the code ? Isn't the code something that's self explanatory and code comments enough to undertstand why a context specefic choice was made ?

.github/skills/ci-efficiency/SKILL.md
HatemMn: https://github.com/Cosmian/kms/pull/7 ? Good ol days 🥲

.github/skills/conventional-commit/SKILL.md
HatemMn: In **my** case this skill will not change anything because I consider that autonomous agents should not touch commits, mine are forbidden from doing it

For people who allow it, I'd say a 1-3 sentences line can summarize all of the skill

.github/skills/kms-test-vector/SKILL.md
HatemMn: Okay amazing, thanks for the doc page !

HatemMn (approved):
Partial review, unblocking merge - Will continue to reply but overall nothing critical to add (didn't read all instructions yet but I beleive they are derivations from what previously was in AGENTS.md )

PR #1018, opened 2026-06-17
Log Call-Site Directory

### LINKED ISSUE #1008: Logging megaDoc
## Context : 

Clients need very precise logging documentation that indicates where **each** log emitted by the KMS is emitted from, categorised by buisness domain or features.

### LINKED ISSUE #1010: MkDocs may break support for all existing plugins and themes soon
**This is a copy paste of what we get once we build the docs**

>  │  ⚠  Warning from the Material for MkDocs team
>  │
>  │  MkDocs 2.0, the underlying framework of Material for MkDocs,
>  │  will introduce backward-incompatible changes, including:
>  │
>  │  × All plugins will stop working – the plugin system has been removed
>  │  × All theme overrides will break – the theming system has been rewritten
>  │  × No migration path exists – existing projects cannot be upgraded
>  │  × Closed contribution model – community members can't report bugs
>  │  × Currently unlicensed – unsuitable for production use
>  │
>  │  Our full analysis:
>  │
>  │  https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/
> 
> ----------------------------------------------------------------
> 
> 
> The owner of MkDocs has completely abandoned maintenance of the project, and instead is planning to publish a "version 2" which will not support
> any existing themes, plugins or even your configuration files. This v2 may eventually replace what you download with `pip install mkdocs`,
> suddenly breaking the build of your existing site.
> 
> To avoid these risks, switch to ProperDocs, a continuation of MkDocs 1.x and a drop-in replacement that supports your current MkDocs setup.
> Simply install it with `pip install properdocs` and build your site with `properdocs build` instead of the MkDocs equivalents.
> 
> Alternatively, to just skip this warning in the future, you can set the environment variable `DISABLE_MKDOCS_2_WARNING=true`.
> 
> For more info visit https://github.com/ProperDocs/properdocs/discussions/33 and https://properdocs.org/
> 
> (This warning was initiated by one of the plugins that you depend on.)

HatemMn:
## Goals: 
- Create a log call site directory database
Rows will be ordered by severity (error → warn → info → debug → trace), then alphabetically by message. Test files are excluded.
The Logs will also be separated by usage domain then by binary if multiple binaries work alltogether within the same usage domain.
The **UI will also count as domain** despite it not being a Rust code.
- Find a way to make the CI ensure that the directory is updated before every release

## How this was done:
- `documentation/docs/configuration/log-reference.md` was first created in a manual way, be doing an extensive regex based grep on all logs at instant t of the codebase. **Note that a lot of the final page content is generated at compile time via mkdocs, including the package version that it gets directly from the project files.**
- A JavaScript filter widget was added to the rendered docs page so operators can narrow by severity badge or keyword without leaving the browser.
- For future code:  Python scanner (`scripts/update_log_index.py`) that extracts log call-sites from Rust and TypeScript source using regex, deduplicates them, resolves `×N` multiplicity markers, and writes  in a fixed table format.

The script ships three modes: **interactive** (prompts user on stale entries), **`--non-interactive`**
(CI/pre-commit auto-apply), and **`--check`** (read-only, exits 1 on any drift; used by CI
and pre-commit hooks).

- As implied by the previous message, adequate CI gate, pre commit hooks, and MISE scripts were added to let this new flow integrate into the existing CI/CD flow

### DIFF
--- a/documentation/hooks.py
+++ b/documentation/hooks.py
@@ -1,0 +1,91 @@
+"""
+MkDocs hooks for the KMS documentation.
+
+on_config:        read KMS version from Cargo.toml → config.extra.kms_version
+on_page_markdown: transform plain file paths in log-reference.md table cells
+                  into GitHub source links at build time.
+"""
+
+import re
+from pathlib import Path
+
+
+# ── on_config ─────────────────────────────────────────────────────────────
+
+def on_config(config):
+    """Called once before the build. Reads version and patches config."""
+    hooks_dir = Path(__file__).parent   # documentation/
+    repo_root  = hooks_dir.parent       # repo root
+    cargo_toml = repo_root / "Cargo.toml"
+
+    version = _read_workspace_version(cargo_toml)
+    config.extra["kms_version"] = version or "develop"
+    return config
+
+
+def _read_workspace_version(cargo_toml: Path) -> str | None:
+    if not cargo_toml.exists():
+        return None
+    text = cargo_toml.read_text(encoding="utf-8")
+    m = re.search(
+        r'\[workspace\.package\].*?^version\s*=\s*"([^"]+)"',
+        text, re.DOTALL | re.MULTILINE,
+    )
+    if m:
+        return m.group(1)
+    m2 = re.search(r'^version\s*=\s*"([^"]+)"', text, re.MULTILINE)
+    return m2.group(1) if m2 else None
+
+
+# ── on_page_markdown ───────────────────────────────────────────────────────
+
+_FILE_CELL_RE = re.compile(r'^`([^`]+\.(rs|ts|tsx|js|toml|py|md|sh))`$')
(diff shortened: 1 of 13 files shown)


Manuthor (commented):
What is the purpose of this script?

documentation/hooks.py
Manuthor, line 1: What is the purpose of this script?

Manuthor (commented):
Great job!

documentation/hooks.py
HatemMn: This script is called at build time (hence the name) and generates via mkdocs the following : 
- The version (that it gets from the cargo toml, avoiding the need for maintainers to manually keep doc in sync)
- The links toward the exact code of the said version (not hard once we have the version, actually)

Manuthor (approved):

PR #1037, opened 2026-07-01
feat: automate key rotation using scheduler

Manuthor:
- add scheduler to monitor key to rotate
- when creating new cert, fix public/private key lifecycle 
- certs benefit now of the keyset feature

### DIFF
--- a/CHANGELOG/feat_automate_key_rotation.md
+++ b/CHANGELOG/feat_automate_key_rotation.md
@@ -1,0 +1,35 @@
+# CHANGELOG — feat/automate_key_rotation
+
+## Features
+
+### Server / Auto-rotation scheduler
+
+- Implement `run_auto_rotation()` in `crate/server/src/core/operations/auto_rotate.rs`:
+  queries `find_due_for_rotation(now)` via the database backend, routes symmetric keys
+  to `ReKey` and private keys to `ReKeyKeyPair`, and skips HSM-resident keys (UID prefix
+  `hsm::`) since the KMS cannot generate new key material inside an HSM ([#970](https://github.com/Cosmian/kms/pull/970))
+- Implement `dispatch_renewal_warnings()`: queries keys due within the next 30 days and
+  emits `info!` log entries at the [30, 7, 1]-day warning thresholds as a skeleton for the
+  forthcoming SMTP notification layer (PR #971) ([#970](https://github.com/Cosmian/kms/pull/970))
+- Add `days_until_next_rotation(attrs, now)` pure function: computes whole days until the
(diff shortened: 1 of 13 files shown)


CHANGELOG/feat_automate_key_rotation.md
```diff
@@ -0,0 +1,35 @@
...
+- Implement `run_auto_rotation()` in `crate/server/src/core/operations/auto_rotate.rs`:
+  queries `find_due_for_rotation(now)` via the database backend, routes symmetric keys
+  to `ReKey` and private keys to `ReKeyKeyPair`, and skips HSM-resident keys (UID prefix
+  `hsm::`) since the KMS cannot generate new key material inside an HSM ([#970](https://github.com/Cosmian/kms/pull/970))
+- Implement `dispatch_renewal_warnings()`: queries keys due within the next 30 days and
+  emits `info!` log entries at the [30, 7, 1]-day warning thresholds as a skeleton for the
+  forthcoming SMTP notification layer (PR #971) ([#970](https://github.com/Cosmian/kms/pull/970))
+- Add `days_until_next_rotation(attrs, now)` pure function: computes whole days until the
+  next rotation deadline using `rotate_date + interval` (if the key has already been rotated
+  at least once) or `initial_date + rotate_offset + interval` (first rotation) ([#970](https://github.com/Cosmian/kms/pull/970))
```
charming-wicket-5502, line 16: Why not use last_rotated date ? Keeping track of offset seems unnecessary

charming-wicket-5502 (approved):

CHANGELOG/feat_automate_key_rotation.md
Manuthor: The code already prefers `rotate_date` (last-rotation date) when it's set — see `days_until_next_rotation` at [auto_rotate.rs:316](https://github.com/Cosmian/kms/blob/b5616a572afa3a78d491d60d3609fc10edf23579/crate/server/src/core/operations/auto_rotate.rs#L316): `if let Some(last_rotate) = attrs.rotate_date` takes precedence.

`rotate_offset` is only used as a fallback for the **first** rotation, when `rotate_date` hasn't been set yet (the key has never been rotated). It's a KMIP 2.1 standard attribute (`Attribute::RotateOffset`) that allows operators to delay the *initial* rotation independently of the recurring interval. For example, a key with a 30-day interval and a 90-day offset will have its first rotation at day 90, then rotate every 30 days thereafter.

Once the first rotation completes, the replacement key gets a `rotate_date`, and from that point on the offset is irrelevant — subsequent rotations use `rotate_date + interval`. So we're not "keeping track of offset" across rotations; it's only a one-time initial delay.

PR #1053, opened 2026-07-10
fix: conf strict parsing

### LINKED ISSUE #1000: Make parsing of TOML configurations strict (server + ckms)
There is a fault-tolerance in both TOML configurations that should be forbidden to avoid false-positive.
If a field is unknown, it must exist early.

Manuthor:
Closes #1000

### DIFF
--- a/.mise/scripts/pykmip/kms.toml
+++ b/.mise/scripts/pykmip/kms.toml
@@ -14,10 +14,6 @@ force_default_username = false
 # the URL should be something like <https://cse.my_domain.com/google_cse>
 # google_cse_kacls_url = "<google cse kacls url>"
 
-# This setting disables the validation of the tokens used by the Google Workspace CSE feature of this server
-# Useful for testing purposes
-google-cse-disable-tokens-validation = false
-
 # This setting enables this server's Microsoft Double Key Encryption service feature.
 # It should contain the external URL of this server as configured in Azure App Registrations
 # as the DKE Service (<https://learn.microsoft.com/en-us/purview/double-key-encryption-setup#register-your-key-store>)
--- a/crate/server/src/config/command_line/clap_config.rs
+++ b/crate/server/src/config/command_line/clap_config.rs
@@ -1193,6 +1193,6 @@ mod tests {
              \"/tmp/kms\"\n";
         let value: toml::Value = toml::from_str(toml_str).unwrap();
         let result = ClapConfig::validate_toml_keys(&value);
-        assert!(result.is_ok(), "should accept all-known keys: {:?}", result);
+        assert!(result.is_ok(), "should accept all-known keys: {result:?}");
     }
 }
--- a/crate/server/src/config/command_line/clap_config.rs
+++ b/crate/server/src/config/command_line/clap_config.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashMap, HashSet},
+    collections::HashMap,
     fmt::{self},
     path::PathBuf,
 };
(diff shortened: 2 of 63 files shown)


crate/server/src/config/command_line/clap_config.rs
```diff
@@ -245,6 +245,84 @@ pub struct ClapConfig {
...
+            "idp_auth",
+            "ui_config",
+            "google_cse_config",
+            "azure_ekm_config",
+            "workspace",
+            "logging",
+            "aws_xks_config",
+            "kmip", // renamed from kmip_policy via #[serde(rename = "kmip")]
+            "jwks_endpoint",
+        ];
```
HatemMn: while not mandatory is it possible to add in the adequate agent file or any dev guideline or even a simple comment of the config object that warns to update this test ? It will eventually fail at any new section added (but it's ok)

.mise/scripts/pykmip/kms.toml
```diff
@@ -14,10 +14,6 @@ force_default_username = false
 # the URL should be something like <https://cse.my_domain.com/google_cse>
 # google_cse_kacls_url = "<google cse kacls url>"
 
-# This setting disables the validation of the tokens used by the Google Workspace CSE feature of this server
-# Useful for testing purposes
-google-cse-disable-tokens-validation = false
-
```
HatemMn, line 20: why ?

crate/server/src/middlewares/otel_http_middleware.rs
```diff
@@ -5,98 +5,46 @@
...
+///
+/// Pass `None` to install a zero-overhead pass-through (metrics disabled).
+///
+/// Place as the **outermost** `App`-level wrap to measure total latency.
+pub(crate) fn otel_http_metrics_middleware<S, B>(
     metrics: Option<Arc<OtelMetrics>>,
-}
-
-impl<S, B> Service<ServiceRequest> for OtelHttpMetricsService<S>
+) -> impl Transform<S, ServiceRequest, Response = ServiceResponse<B>, Error = Error, InitError = ()>
```
HatemMn, line 29: Why this update ?

crate/server/src/routes/google_cse/operations.rs
```diff
@@ -1178,7 +1178,7 @@ async fn cse_wrapped_key_decrypt(
         authenticated_encryption_additional_data: resource_name,
         authenticated_encryption_tag: Some(authenticated_tag.to_vec()),
     };
-    let key = Box::pin(decrypt(kms, decryption_request, &user)).await?;
```
HatemMn, line 1181: why ?

HatemMn (approved):
red the whole PR but seems to have gone way beyond the scope of the title ? Is it possible to clarity the goals in the description ?

To my understanding : 
- the toml issues are cleard
- toolchain updated
- some middlewares were updated and code reduced

LGTM anyway !

crate/server/src/config/command_line/clap_config.rs
Manuthor: Superseded by a better fix: replaced `validate_toml_keys` and the static `ALLOWED` list with [`serde_ignored`](https://crates.io/crates/serde_ignored) (by @dtolnay). The crate wraps the serde deserializer and reports any unknown field via callback — no manual list to maintain. Works for top-level fields including those injected via `#[serde(flatten)]` (e.g. `HsmConfig`) where `deny_unknown_fields` is incompatible with flatten. New sections are detected automatically.

.mise/scripts/pykmip/kms.toml
Manuthor: The key `google-cse-disable-tokens-validation` (hyphens) at the top level was always silently ignored: the actual field lives under `[google_cse_config]` with underscores (`google_cse_disable_tokens_validation`). With strict unknown-field detection now enforced, this top-level unknown key would cause a startup error. Since it was both at the wrong nesting level and used the wrong key syntax, it had no effect. Safe to remove — the field defaults to `false`.

crate/server/src/middlewares/otel_http_middleware.rs
Manuthor: Opportunistic cleanup. The manual `Transform` + `Service` pair (~90 lines of boilerplate) was replaced with [`actix_web::middleware::from_fn`](https://docs.rs/actix-web/latest/actix_web/middleware/fn.from_fn.html) — the idiomatic pattern for stateful pass-through middleware in actix-web 4.x. Identical behaviour, ~70 fewer lines.

crate/server/src/routes/google_cse/operations.rs
Manuthor: `Box::pin(future).await` forces the future onto the heap before polling, adding a heap allocation per call. It is needed when the future must be type-erased (`dyn Future`) or when the calling function is recursive (which would produce an infinite-size future type). Neither applies here: `decrypt` is a plain non-recursive `async fn` and its result is consumed immediately. The removal eliminates an unnecessary allocation with no behavioural change.

Issue #1056, opened 2026-07-10
Requests sometimes fail unexpectedly.


AdamJSoftware:
When using the app, I have notice a lot of the following failure:
```
err=HTTP Client: HTTP request failed: client error (Connect)
```

This usually happens in the following code.

```rust
            let kek = kms_client
                .create(cosmian_kms_client::kmip_2_1::kmip_operations::Create {
                    object_type: ObjectType::SymmetricKey,
                    attributes: cosmian_kms_client::kmip_2_1::kmip_attributes::Attributes {
                        cryptographic_algorithm: Some(CryptographicAlgorithm::AES),
                        ..Default::default()
                    },
                    protection_storage_masks: None,
                })
                .await?;
            println!("kek created");

            let kek_id = kek.unique_identifier;

            let active_model = ActiveModel {
                kms_key_id: ActiveValue::Set(kek_id.to_string()),
                // version: ActiveValue::Set(1),
                organization_id: ActiveValue::Set(organization_id.clone()),
                ..Default::default()
            };
            let model = active_model.insert(db).await.map(|f| f.into())?;
            kms_client.activate(Activate {
                unique_identifier: kek_id
            }).await?;

            println!("kek activated");
            Ok(model)
```

The KmsClient is long lived, not sure if that has anything to do with it,

AdamJSoftware:
seems to be happening non-deterministcally for any operation involving the kms, most of the time it works, other times, it will first fail, then it will work right after.

AdamJSoftware:
Seem like a connection is dropped which causes the request to fail:

```
2026-07-13T00:26:06.864240Z TRACE hyper_util::client::legacy::pool: checkout waiting for idle connection: ("https", localhost:9998)
2026-07-13T00:26:06.866603Z TRACE hyper_util::client::legacy::pool: checkout dropped for ("https", localhost:9998)
HttpClient(
    "HTTP request failed: client error (Connect)",
)
None
2026-07-13T00:26:08.698444Z TRACE hyper_util::client::legacy::pool: checkout waiting for idle connection: ("https", localhost:9998)
2026-07-13T00:26:08.715169Z TRACE hyper_util::client::legacy::client: http1 handshake complete, spawning background dispatcher task
2026-07-13T00:26:08.715233Z TRACE hyper_util::client::legacy::client: waiting for connection to be ready
2026-07-13T00:26:08.715312Z TRACE hyper_util::client::legacy::client: connection is ready
[...]

PR #1061, opened 2026-07-14
fix: stale connections in ckms

Manuthor:


crate/clients/client/src/http_client/client.rs
```diff
@@ -528,13 +528,72 @@ impl HttpClient {
...
+    }
 
+    /// Decompose `request` and clone it into two identical instances.
+    ///
+    /// The second one is boxed so that the caller only has to store a pointer in
+    /// its async state machine (avoids inflating future sizes).
+    // The return type encodes a specific ownership pattern (one owned + one boxed) needed
+    // to keep the async state machine small; a type alias would obscure that intent.
+    #[allow(clippy::type_complexity)]
+    fn prepare_requests(
```
HatemMn: Does this copy **all** requests ? Assuming that the bug happens, let's 1% of the time, this seems very sub-otimal and will double connection handling overhead

I will write another comment to explain my idea better

crate/clients/client/src/http_client/client.rs
```diff
@@ -528,13 +528,72 @@ impl HttpClient {
     }
```
HatemMn, line 540: Alternative fix: I propose the following idea instead of letting the problem happen in the first place

adding a `.pool_timer(TokioTimer::new())` + a `pool_idle_timeout` slightly below the server's actix keep-alive (default 5s ? I think 120s for ours) ==> Fewer stale conns accumulate ==> Reduces retries, **doesn't remove the race** - so keep the current fix as correctness garantee

This way, we will gain in computation time: problem will be more rare

Here is an AI generated plan that describes a way to do it

```
Steps

Add TokioTimer import in client.rs — change the existing use hyper_util::{client::legacy::Client, rt::TokioExecutor}; to also import TokioTimer from hyper_util::rt.

Wire the timer + timeout into the builder at the single Client::builder(TokioExecutor::new()).build(...) call (line 366). Replace it with:


Client::builder(TokioExecutor::new())    .pool_idle_timeout(Duration::from_secs(90))   // < server's 120 s    .pool_timer(TokioTimer::new())    .build(https_connector)
The Duration import is already available via std::time::Duration — add it to the use list if not already present.

No Cargo.toml changes needed — hyper-util already declares the tokio feature (line 36–41 of Cargo.toml), which gates TokioTimer.
```

crate/clients/client/src/http_client/client.rs
```diff
@@ -528,13 +528,72 @@ impl HttpClient {
...
+            for (name, value) in &p.headers {
+                builder = builder.header(name, value);
+            }
+            builder
+                .body(b.clone())
+                .map_err(|e| HttpClientError::Default(format!("Failed to build request: {e}")))
+        };
+
+        Ok((build(&parts, &body)?, Box::new(build(&parts, &body)?)))
+    }
```
HatemMn, line 596: I propose to delete this function entirely, instead : 

- Decompose the request in (parts,body)
- Build first_req from `http::Request::from_parts(parts.clone(), body.clone())` ,  this is O(1)
- On Ok -> move on
- On other errors -> move on
- On **Err(e) if e.is_connect()** build retry inline

crate/clients/client/src/http_client/tests.rs
```diff
@@ -695,3 +695,122 @@ mod custom_headers_tests {
...
+            successes, N,
+            "All {N} requests must succeed; {successes} succeeded, server accepted \
+             {connections} connections"
+        );
+        // Each request must have resulted in a new TCP connection because the
+        // server closes every connection after responding.
+        assert_eq!(
+            connections, N,
+            "Expected {N} TCP connections (one per request), got {connections}"
+        );
```
HatemMn: isn't this a server behavior check ? (which is off topic to the client developed here)

HatemMn (changes requested):
good solution, slightly overengibneered but overall that's the fix

It might be interesting to source against an issue ([like this one](https://github.com/hyperium/hyper/issues/2136) ) that does confirm that "this" is the solution, as it looks a bit counter intuitive at first reading

AGENTS.md
```diff
@@ -291,7 +291,18 @@ Run **`/kms-changelog`** — it reads the branch name, determines the correct fi
...
+- Bug fix (any user- or operator-observable error, crash, or wrong behaviour)
+- New feature (user-visible or operator-visible)
+- Security hardening or vulnerability fix
+- Behavioral change (default behaviour, error messages, CLI output)
+- Breaking change (API signatures, CLI flags, config keys, algorithms)
+- Significant refactor that materially affects operators or contributors
+
+**Do not** write an entry for: routine internal implementation, formatting, test-only changes with no observable behaviour difference, CI pipeline adjustments.
+
+> **Tip for agents**: If this step was not completed in the current session, the entry must still be created before the PR is merged. Check `CHANGELOG/<branch>.md` exists; if not, create it now.
```
HatemMn: two points about this : 

- Are internal bugs to be referenced in hight level changelogs
- Are all these necessary ? The overall idea is reducing agent MD in a cupid manner - I Think all these bullets fit in 1-2 sentences ? And the agent can be guided to go check the skill for "Details) ( I see the skill is already created, which justifies this should be kep short)

crate/clients/client/src/http_client/client.rs
Manuthor: This concern is addressed by the `prepare_requests` design (see the reply on the next thread). The upfront clone is cheap in practice: `Bytes` is reference-counted (O(1) clone), `Method` is a small enum, `Uri` and `HeaderMap` are proportional to header count — typically a handful of headers. The real trade-off is discussed inline.

crate/clients/client/src/http_client/client.rs
Manuthor: Fixed. Added `pool_idle_timeout(Duration::from_secs(90))` + `pool_timer(TokioTimer::new())` to the `Client::builder` call (commit 0f932095). This evicts idle connections before the server's keep-alive window (Actix-web default 120 s), reducing the frequency of the stale-connection race. `TokioTimer` was already available via the `tokio` feature of `hyper-util` declared in `Cargo.toml`. The `is_connect()` retry is kept as the correctness guarantee for the residual race window.

crate/clients/client/src/http_client/client.rs
Manuthor: Not applicable as described. Storing `parts: http::request::Parts` and `body: Full<Bytes>` as live variables across the `self.client.request(req).await` point inflates the async state machine and causes `clippy::large_futures` warnings in every caller in `cosmian_kms_cli_actions` (confirmed during development — the google, aws, and azure BYOK callers all triggered it). `prepare_requests` is the minimal encapsulation that moves both request copies out of the state machine synchronously before any `.await`, and boxing the retry reduces the cross-await storage to 8 bytes (a pointer). The concern about always paying the clone cost is valid, but the clone is cheap (Bytes is ref-counted, headers are small), and the alternative would require reintroducing the lint.

crate/clients/client/src/http_client/tests.rs
Manuthor: Not applicable. The TCP server is a test fixture; the system under test is `HttpClient.get()`. The test documents the observable contract: a long-lived `HttpClient` must succeed even when the underlying server closes connections after each response — guarding against accidental removal of the retry logic. One honest caveat: on loopback, TCP delivers FIN and response data in the same segment so hyper rarely pools a stale connection; the test is primarily a documentation-level regression guard. The real reproducer requires a Docker or remote network (exactly the reporter's setup in #1056). Happy to add a comment to the test clarifying this if useful.

AGENTS.md
Manuthor: Fixed (commit 0f932095). Both points addressed: (1) The entry in `CHANGELOG/fix_stale_connections_in_ckms.md` covers a user-reported production error (#1056), so it belongs in the changelog. (2) AGENTS.md step 7 has been shortened to two sentences pointing at the `/kms-changelog` skill for the full criteria — the detail belongs in the skill, not in the top-level instructions.

crate/clients/client/src/http_client/client.rs
HatemMn: Sorry to reopen this but before moving on can you reduce the comment verbosity ? A sentence (or even nothing) is enough I think

        // Build hyper client with connection pooling.
        // `pool_idle_timeout` is set below the server's keep-alive window (120 s default
        // for Actix-web) so idle connections are evicted before the server closes them,
        // reducing the frequency of stale-connection races.  The `is_connect()` retry in
        // `send()` handles any residual race that slips through.

crate/clients/client/src/http_client/tests.rs
HatemMn, line 1: Follow up to this : https://github.com/Cosmian/kms/pull/1061/changes#r3586753278

After re thinking, I re re-reaise this point again and I think that let's just delete this test or replace it with some E2E test that forces the retry branch to actually work out

First I tried an empirical approach : 
1. Patched send() to strip retry arm entirely (simulate pre-fix code), rebuilt, ran test 50 times with some shell loop
2. `PASS=50 FAIL=0`

**Conclusion: This test will pass wherever the retry path exists or not**

I think that so by the time loop iteration N+1 calls .get(), pool already evicted the dead conn. The fresh connect happens because pool is empty, not because a stale conn got reused and retry saved it.
Test holds true regarless of what's happenning

Also, mocking a server feels like an overkill for such a low importance bug, we should rather mock the KMS or use the usual test_kms_server with the right config

crate/clients/client/src/http_client/client.rs
HatemMn: Maybe I didn't understand this comment was a reply of another question ? I have not commentted about clippy lint (It's fine) but rather proposed to reduce code size by
1) deleting a function
2) putting its inside flow within the adequate error branch to **not process every request** this way

I will leave up to you the details and trust your remarks about its cheapness since i wasn't the one who dev it

crate/clients/client/src/http_client/client.rs
```diff
@@ -362,8 +367,15 @@ impl HttpClient {
...
 
-        // Build hyper client with connection pooling
-        let client = Client::builder(TokioExecutor::new()).build(https_connector);
+        // Build hyper client with connection pooling.
+        // `pool_idle_timeout` is set below the server's keep-alive window (120 s default
+        // for Actix-web) so idle connections are evicted before the server closes them,
+        // reducing the frequency of stale-connection races.  The `is_connect()` retry in
+        // `send()` handles any residual race that slips through.
+        let client = Client::builder(TokioExecutor::new())
+            .pool_idle_timeout(Duration::from_secs(90))
```
HatemMn: nitpick: who garantees that this duraction will be always inferior to the other duration ? Is a refactor (as constants) something we can consider ? And this way this magic value becomes (forever) dependent on server keep alive window

Comment says 120s is actix default, this has no regards to our codebase and might not be the case at all - I doubt operators might change that value (didn't check)

crate/clients/client/src/http_client/client.rs
```diff
@@ -362,8 +367,15 @@ impl HttpClient {
                 HttpClientError::Default(format!("Failed to build HTTPS connector: {e}"))
             })?;
 
-        // Build hyper client with connection pooling
-        let client = Client::builder(TokioExecutor::new()).build(https_connector);
+        // Build hyper client with connection pooling.
+        // `pool_idle_timeout` is set below the server's keep-alive window (120 s default
+        // for Actix-web) so idle connections are evicted before the server closes them,
+        // reducing the frequency of stale-connection races.  The `is_connect()` retry in
+        // `send()` handles any residual race that slips through.
```
HatemMn: Sorry to reopen this but before moving on can you reduce the comment verbosity ? A sentence (or even nothing) is enough I think

        // Build hyper client with connection pooling.
        // `pool_idle_timeout` is set below the server's keep-alive window (120 s default
        // for Actix-web) so idle connections are evicted before the server closes them,
        // reducing the frequency of stale-connection races.  The `is_connect()` retry in
        // `send()` handles any residual race that slips through.

HatemMn (changes requested):
Thanks for the update ! detected a bug within the test and some refactor idea

Small bug, smaller code, big PR :')

crate/clients/client/src/http_client/client.rs
Manuthor: Fixed (commit 052c2f9b). Reduced to one line: `// Evict idle connections before the server's keep-alive window closes them.`

crate/clients/client/src/http_client/tests.rs
Manuthor: Fixed (commit 052c2f9b). The test is deleted — you are right that it passed regardless of whether the retry path existed (the pool eviction already ensured a fresh connect). Agreed that an E2E test exercising the actual retry path would be the right investment if needed in the future.

crate/clients/client/src/http_client/client.rs
Manuthor: Fixed (commit 052c2f9b). Replaced `prepare_requests` with `split_for_retry` — simpler body (no inner closure, uses method/uri/headers accessors directly), no complex ownership dance. The Box is still necessary: without it the async state machine grows by ~165 bytes per `send()` invocation and `clippy::large_futures` fires on several callers (confirmed during implementation). `#[expect(clippy::type_complexity)]` with an inline reason replaces the old `#[allow]`.

crate/clients/client/src/http_client/client.rs
Manuthor: Fixed (commit 052c2f9b). Added `HttpClient::POOL_IDLE_TIMEOUT_SECS: u64 = 90` with a doc comment explaining it must stay below the server's keep-alive window (120 s, configured in `start_kms_server.rs`). The constant is used in the builder call, making the relationship explicit and easy to update in one place.

crate/clients/client/src/http_client/client.rs
Manuthor: Fixed (commit 052c2f9b) — same change as the other thread.

crate/clients/client/src/http_client/tests.rs
Manuthor: The stale-connection test has now been removed (commit 052c2f9b) per the follow-up review. Agreed with the empirical finding.

HatemMn (approved):
Looks way better !

PR #1082, opened 2026-07-31
Spire fix b4 ma

HatemMn:
**_(This description was generated using some LLM)_**

Here are the four changed files, one by one:

**1. spire**
The repo-native integration test harness for the whole SPIRE flow (non-FIPS only). It boots the auth-verifier (port 8443) and KMS (port 9998), wipes stale state (`/tmp/auth-verifier-spire-test.db`, stray port listeners), spins up the multi-tenant SPIRE + Mistral-client stack via `docker compose --profile spire`, then runs the conformance + adversarial scenarios (cross-tenant isolation, path traversal, secret_id races, log gate, etc.). It reads all config from spire — no personal/.agents paths. This is the team's `mise run test:spire` entry point.

**2. spire_token.rs**
The app-token validation middleware for the SPIRE-compatible API. It reads the `X-Vault-Token` header, hashes it with SHA-256, and checks a bounded 30-second in-memory cache (`SpireTokenCache`, capped at 100k entries with opportunistic sweeping). On a cache miss it calls `GET /auth/token/lookup-self` on the auth-verifier, then injects both `SpireAuthenticatedUser` and `AuthenticatedUser` into the request extensions so the KMS permission system attributes the request to the real caller instead of `default_username`.

**3. auth_proxy.rs**
A transparent reverse proxy for the `/v1/auth/*` scope. When `vault_api_enabled = true` the KMS becomes SPIRE's single `vault_addr`, and this forwards AppRole login / `lookup-self` / `renew-self` / `revoke-self` to the auth-verifier (`vault_auth_verifier_url`), stripping the `/v1` prefix. It preserves method, path, query, `X-Vault-Token`/`Content-Type`/`Authorization`, and body; returns the upstream status/body byte-for-byte; returns 502 if the auth-verifier is unreachable. It also rejects `.`/`..` path segments up front to block traversal out of the `/auth/*` scope (the B-4 fix).

**4. start_kms_server.rs**
The server bootstrap. It builds the Actix `App` and, when the Vault API is enabled at runtime, registers the SPIRE scopes — the `/v1/auth` proxy scope (wired to `proxy_auth_request` with the shared `reqwest::Client` + auth-verifier `Url` in `app_data`), the transit scope (with `spire_token_middleware` and `JsonConfig::content_type_required(false)`), and PKI. This is where the middleware and proxy handler from files 2 and 3 get mounted.


Manuthor:
Closing since git cherry-pick in 5b241a99

Issue #937, opened 2026-04-29
Phase 2: SIEM export (the integration layer)


HatemMn:
Once Phase 1 is on to production, we can move to integrating with SIEMs and audit systems

In a nutshell : 

Goal: OTel, syslog/CEF, Splunk HEC exporters. They read from the same local audit event stream (either by subscribing to new events or polling. TB thought)

Whatever integration should be **Entirely optional** if no [audit.otlp] / [audit.syslog] / [audit.splunk_hec] is configured, Phase 1 still works fully

**The local store is the source of truth. External SIEMs are consumers. That separation also means: if a SIEM is misconfigured or down, you don't lose audit events — they're already persisted locally and can be replayed.**

## Useful ressource : 

https://github.com/vectordotdev/vector

PR #1115, opened 2026-08-11
test: add SIEM integration tests

Manuthor:
Proves issue #937 (syslog/CEF, SIEM integration):

- mise test:audit  — tamper-evident JSONL, hash chain, fields
- mise test:cef    — CEF v27 format (jc), UDP syslog, TCP rsyslog
- mise test:siem   — fluent-bit, filebeat
- mise test:monitoring — OTel+VictoriaMetrics+Grafana stack health


documentation/docs/configuration/otlp-telemetry.md
HatemMn, line 1: I think this file is stale documentation from the part that was deleted, shall be deleted too ?

.mise/tasks/test/audit
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+#MISE description="Run audit log integration test (ADR-003/004: JSONL, hash chain, fields)"
+#USAGE flag "-v --variant <variant>" env="VARIANT" help="Feature variant (must be non-fips)" default="non-fips" {
+#USAGE   choices "fips" "non-fips"
+#USAGE }
+# Proves ADR-003 (tamper-evident JSONL single-writer) + ADR-004 (HTTP audit middleware).
+# Requires non-fips: audit middleware and CEF features are only in non-fips builds.
+# No Nix shell needed. Uses Cargo to build KMS binary.
```
HatemMn, line 8: Audit is not gated behind non-fips, why this constraint ?

lychee.toml
HatemMn, line 1: Why this file is related to this PR ?

documentation/nav.yml
```diff
@@ -132,15 +132,15 @@ nav:
       - Authorizing users with access rights: configuration/authorization.md
       - Enabling TLS: configuration/tls.md
       - Obtaining TLS Certificates: configuration/certificates.md
-      - Logging and telemetry: configuration/logging.md
-      - Log reference: configuration/log-reference.md
+      - Observability:
+          - Log output: configuration/logging.md
+          - Log call-site reference: configuration/log-reference.md
+          - Metrics & traces (OTLP): configuration/otlp-telemetry.md
+          - Monitoring stack (Grafana): configuration/monitoring-setup.md
```
HatemMn: To my understanding, this is a doc change related to the Observability stack, maybe this should be deferred to its own PR to avoid errors ?

For example, if logg call-site directory is decided to now become part of the root tree, this comment should be updated to reflect it : 

Image

HatemMn: Other thing to note especially if moving this : duckduckgo throws a 404 when someone searches this on google:

https://docs.cosmian.com/key_management_system/configuration/log-reference/

Image

.mise/scripts/test/test_siem_cef_tcp_syslog.sh
```diff
@@ -0,0 +1,307 @@
...
+# The message includes the syslog PRI header + CEF line.
+# rsyslog's imtcp module with default framing supports this format.
+
+sent=0
+while IFS= read -r line; do
+  [[ "${line}" == CEF:* ]] || continue
+  # Assemble a minimal RFC 3164 syslog message (PRI 134 = local0.info)
+  syslog_msg="<134>$(date '+%b %d %H:%M:%S') kms-audit: ${line}"
+  # RFC 6587 octet-counting: space between count and message, count = len(message)
+  printf '%zu %s' "${#syslog_msg}" "${syslog_msg}" |
```
HatemMn, line 220: `${#msg}` is a bash character count, not a byte (or octet) count

in other terms, any non ASCII character in the name field will produce a wrong count, producing an error that might cascade

.mise/scripts/test/test_siem_cef_tcp_syslog.sh
```diff
@@ -0,0 +1,307 @@
...
+  # Verify required CEF extension keys are present
+  for key in rt= suser= outcome= act= cn1= cn1Label=durationMs externalId=; do
+    if ! echo "${cef_line}" | grep -q "${key}"; then
+      echo "FAIL: missing CEF extension key '${key}' in: ${cef_line}" >&2
+      cef_fail=$((cef_fail + 1))
+      continue 2
+    fi
+  done
+  cef_ok=$((cef_ok + 1))
+done <"${VERIFY_CEF}"
```
HatemMn, line 293: While the comment does say to use the python helper : 
> # Use the Python helper to validate CEF structure.

Actually it's not used down here and some ADHOC bash is used

In contrast, the use files ( like `test_siem_filebeat.sh line 373` does call the python script: `python3 "${HELPER}" es "${VERIFY_JSONL}" --exact "${audit_lines}"` )

documentation/docs/configuration/siems.md
```diff
@@ -11,17 +11,57 @@ specification reference — see [CEF export format](./cef-export.md).
...
+
+```mermaid
+sequenceDiagram
+    participant File as audit.jsonl
+    participant ckms as ckms audit export
+    participant Syslog as syslog listener
+
+    Note over File,ckms: Run on demand or via cron
+    ckms->>File: read JSONL (optional --since / --until range)
+    ckms->>Syslog: CEF lines on stdout
```
HatemMn, line 59: Something I fear with this diagram is that it might mislead readers into understanding something else

The CLI here is shown as a standalone actor, meanwhile, it's just a CLI that reads, formats, and prints back to stdout. 

To my understanding there is some separate shell loop that actually puts bytes on the wire. ckms has no network sink of its own.

Someone who reads only the diagram walks away thinking the tool ships events directly to a syslog endpoint, which would be a wrong mental model

Possible fixes:
- easy fix: deleting the diagram, keeping written docs (it's pretty trivial to understand)
- if kept: adding a third lane/box/actor/something (ckms -> shell/nc -> Syslog) to match what the code examples actually do.

.mise/scripts/test/test_siem_cef_tcp_syslog.sh
HatemMn: Also worth checking if the ADRs do follow on this

.github/workflows/test_all.yml
```diff
@@ -85,6 +88,17 @@ jobs:
...
+          # run non-fips only to avoid duplicating test time.
+          - type: siem
+            features: fips
+          # audit integration tests require non-fips (audit middleware + CEF features)
+          - type: audit
+            features: fips
+          # monitoring tests work with both variants, but run non-fips only to avoid
+          # duplicating Docker-heavy test time (OTel + VictoriaMetrics + Grafana).
+          - type: monitoring
+            features: fips
```
HatemMn: should be differed to OTEL PR ?

.github/workflows/test_all.yml
```diff
@@ -85,6 +88,17 @@ jobs:
             features: fips
           - type: cef
             features: fips
+          # siem integration tests are format-agnostic (same audit output in both variants);
+          # run non-fips only to avoid duplicating test time.
+          - type: siem
+            features: fips
+          # audit integration tests require non-fips (audit middleware + CEF features)
+          - type: audit
+            features: fips
```
HatemMn, line 96: No non-fips gate requires the non-fips build, which kindof loses us some free coverage

documentation/docs/configuration/siems.md
```diff
@@ -124,7 +216,38 @@ The `ckms audit export --format cef` command converts the JSONL audit store into
...
+Example with rsyslog — add to `/etc/rsyslog.conf` or `/etc/rsyslog.d/kms.conf`:
+
+```conf
+module(load="imtcp")
+input(type="imtcp" port="5514" Ruleset="kms_cef")
+ruleset(name="kms_cef") {
+  action(type="omfile" file="/var/log/kms-cef.log" template="RSYSLOG_ForwardFormat")
+}
+```
+
```
HatemMn, line 233: **This contradicts what is said on the ADR document**

( _copy pasted:_ 
`RSYSLOG_FileFormat` is used (not `RSYSLOG_ForwardFormat`) because it appends `\n`
after each message, producing one line per event — essential for downstream parsing.
)

HatemMn: The example on the ADR is actually tested, and hence more viable

documentation/docs/configuration/audit-logs.md
```diff
@@ -7,6 +7,25 @@ can be parsed by any standard tooling.
...
+    Note over MW: Extracts user, IP, request_id from request
+    MW->>MW: execute KMIP operation
+    MW->>File: append JSON event with row_hash
+    Note over File: row_hash = SHA-256(prev_hash + event fields)
+    MW-->>Client: KMIP response
+    Note over File: Hash chain verified offline with ckms audit verify
+    Agent->>File: tail from last offset
+    Agent->>SIEM: forward events (guaranteed, resumable)
+```
+
```
HatemMn: I beleive this diagram is not in the right place, talking about Middleware ( `Note over MW: Extracts user, IP, request_id from request` ) has close to no value for anyone who the goal isn't to write Rust code

I propose to differ this:
- In Rust comments
- In PR description
- In some internal docs

⚠️ **IMPORTANT**: if ever this stays here please mark me a comment in this PR : https://github.com/Cosmian/kms/pull/1121 about it, because the new AuditSink trait will unvalidate some part of this diagram. Also, if it's kept, Agents should be taken away from it (they're external actors from the KMS)

documentation/docs/configuration/audit-logs.md
```diff
@@ -48,6 +67,34 @@ When `audit.file.path` is omitted the file defaults to `<root-data-path>/audit.j
...
+The hash-chained JSONL file is the sole authoritative source of truth.  An in-process
+channel is best-effort: if the process crashes, the channel buffer is lost; if the
+receiver is down, events are silently dropped with no way to detect the gap.  A
+file-tailing agent reading from its last committed offset survives restarts and recovers
+automatically — delivering every event exactly once.
+
+Per **PCI-DSS Req. 10**, access to the audit trail must be restricted to personnel with a
+documented business need (the SOC).  Route the agent's output to a dedicated, access-controlled
+SIEM pipeline, separate from the application metrics/traces pipeline (operated by SRE).
```
HatemMn: This too specific remark is not really in its place I think

If wanting to keep it: please differ it to the best practices section. Explain the practice first. Mention PCI-DSS as "bonus" by the end of the tip - it's only a remark for operators and this doc is about a KMS's usage

documentation/docs/configuration/audit-logs.md
```diff
@@ -48,6 +67,34 @@ When `audit.file.path` is omitted the file defaults to `<root-data-path>/audit.j
...
+
+See [SIEMs](./siems.md) for end-to-end configuration examples with Vector, Filebeat, and
+Fluent Bit.
+
```
HatemMn: Since now we have this new file ( [SIEM.MD](https://github.com/Cosmian/kms/pull/1121) ) why mix topics here ? Maybe just add a line and redirect those info there - minding to avoid redundancy

documentation/docs/configuration/audit-logs.md
```diff
@@ -48,6 +67,34 @@ When `audit.file.path` is omitted the file defaults to `<root-data-path>/audit.j
...
+channel is best-effort: if the process crashes, the channel buffer is lost; if the
+receiver is down, events are silently dropped with no way to detect the gap.  A
+file-tailing agent reading from its last committed offset survives restarts and recovers
+automatically — delivering every event exactly once.
+
+Per **PCI-DSS Req. 10**, access to the audit trail must be restricted to personnel with a
+documented business need (the SOC).  Route the agent's output to a dedicated, access-controlled
+SIEM pipeline, separate from the application metrics/traces pipeline (operated by SRE).
+
+See [SIEMs](./siems.md) for end-to-end configuration examples with Vector, Filebeat, and
```
HatemMn: There is no end-to-end configarutions provided (at least so far)

.mise/scripts/test/otlp-audit-collector-config.yaml
HatemMn, line 1: I think all this file is an orphan that should be deleted ?

.mise/scripts/test/test_audit_log.sh
HatemMn, line 1: I think this whole test has a big flaw, as it does not prove it's top claime

The test says : 

> # Proves ADR-2026-07-09-audit-log-single-writer-design (tamper-evident JSONL single-writer) and ADR-2026-07-09-audit-middleware-extension-injection (HTTP middleware):

However the only place where something is verified is around here ( see below ) 

```
130: # ── GUARD 2: hash chain must be valid ─────────────────────────────────────────
131:
132: echo "==> Running ckms audit verify --path ${AUDIT_JSONL}..."
133: if ! "${ckms_bin}" audit verify --path "${AUDIT_JSONL}" 2>&1; then
134:   echo "ERROR: ckms audit verify failed — tamper-evident hash chain is broken." >&2
135:   echo "       This proves ADR-2026-07-09-audit-log-single-writer-design hash chain is NOT working correctly." >&2
136:   exit 1
137: fi
138: echo "GUARD OK: hash chain verified (${audit_lines} rows, SHA-256 chain intact)."
```

This tests the happy path, but **does not test tamper evident jsonL**

An actual test that proves that must for example tamper with the logs, flip one random bit and assert verification fails

for info: this has been already tested - but if it's not around I might have not committed it

documentation/docs/SUMMARY.md
```diff
@@ -93,16 +93,15 @@
...
     - [Enabling TLS](configuration/tls.md)
     - [Obtaining TLS Certificates](configuration/certificates.md)
-    - [Logging and telemetry]()
-        - [Logging](configuration/logging.md)
-        - [Log reference](configuration/log-reference.md)
+    - [Observability]()
+        - [Log output](configuration/logging.md)
+        - [Log call-site reference](configuration/log-reference.md)
+        - [Metrics & traces (OTLP)](configuration/otlp-telemetry.md)
+        - [Monitoring stack (Grafana)](configuration/monitoring-setup.md)
```
HatemMn: As agreed all unrealed otel edits must move to another PR, this comment is for the follow up

.mise/tasks/test/siem
HatemMn, line 1: Note from a live test: this file requires requires docker compose up -d elasticsearch as a manual precondition; the task doesn't auto-start it or fail fast — it burns a full timeout (~120s) before printing the fix on first run

documentation/docs/adr/2026-07-09-audit-log-single-writer-design.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "ADR-0003: Tamper-Evident JSONL Audit Log — Single-Writer Architecture"
+title: "ADR-2026-07-09-audit-log-single-writer-design: Tamper-Evident JSONL Audit Log — Single-Writer Architecture"
```
HatemMn: ADR-reference drift: .mise/tasks/test/{audit,siem} headers still cite stale numeric "ADR-003/004" (superseded by the date-based rename in this same PR); separately, .mise/tasks/test/{cef,cef-tcp-syslog} cite "ADR-007" — a number that was never assigned to anything (the TCP-syslog ADR was authored post-rename with no numeric ID at all). Two different failure modes (stale vs. fabricated), same low-effort fix (use the date-based names).

docker-compose.yml
```diff
@@ -153,6 +153,42 @@ services:
...
+      - ES_JAVA_OPTS=-Xms512m -Xmx512m
+    ports:
+      - ${KMS_SLOT_ES_PORT:-9200}:9200
+    healthcheck:
+      test:
+        - CMD-SHELL
+        - curl -sf http://localhost:9200/_cluster/health || exit 1
+      interval: 10s
+      timeout: 5s
+      retries: 12
```
HatemMn, line 190: While noted this is clearly "for tests", someone copy pasting this might enable this in production

Also, I think that this file (root of the whole repo) has deviated from its original purpose

Some time ago, it used to be a very simple docker file to instanciate what the KMS might miss (a redis, a postgres, a mysql DB)

Now, it's a huge file that contains some test infra, spire actors, an insecure sink, isn't this a big problem ? No one will read it, and an innocent use running docker-compose up to get redis will get spammed by unwanted containers

Worth opening an issue about this matter IMO

.mise/tasks/test/siem-filebeat
HatemMn, line 1: test_siem_filebeat.sh misdiagnoses an Elasticsearch disk-watermark shard-allocation failure as "mapping conflict" in its own error message. Confirmed live: root cause was the host's disk at 99%, not the PR's pipeline/mapping (both were verified correct via manual repro). Script never queries _cluster/health/_cluster/allocation/explain before asserting this specific wrong hypothesis — would misdirect a CI triager on any disk-constrained runner.

(PS: this comment is the result of an LLM, I'm not quite sure about it... Close it if you consider it wrong, but the LLM did actually do the execution so it's grounded by something true)

.mise/tasks/test/monitoring
HatemMn, line 1: Seems like this is also from the batch of diffs that should be moved to the observability PR

.mise/scripts/README.md
```diff
@@ -1,961 +1,303 @@
...
+
+### Database backends
+
+| Task | Default variant | Description |
+|------|-----------------|-------------|
+| `test:sqlite` | fips | SQLite workspace tests |
+| `test:psql` | fips | PostgreSQL tests |
+| `test:mysql` | fips | MySQL tests |
+| `test:percona` | fips | Percona XtraDB tests |
+| `test:mariadb` | fips | MariaDB tests |
```
HatemMn, line 43: why did all this change ?

documentation/docs/configuration/siems.md
```diff
@@ -94,7 +184,9 @@ field extraction, facets, or dashboards:
 | `object_uid` | string | Yes | Target object identifier (KMIP `UniqueIdentifier`) |
 | `algorithm` | string | Yes | Cryptographic algorithm (e.g. `AES`, `RSA`) |
 | `client_ip` | string | Yes | Source IP address |
-| `result` | `"Success"` or `{"Failure": "..."}` | No | Outcome — split into `outcome` + `reason` fields |
+| `result` | `"Success"` or `{"Failure": "..."}` | No | Polymorphic — normalize with the ingest pipeline described above into `result_status` + `result_error` |
+| `result_status` | `"Success"` or `"Failure"` | No | Normalized outcome (after pipeline); use for facets and alerts |
+| `result_error` | string | Yes | Normalized error message (after pipeline); present only on Failure events |
 | `duration_ms` | integer | No | Operation latency |
 | `request_id` | string (UUID) | Yes | Correlation ID across batch operations |
```
HatemMn, line 191: This table lists `request_id` as a real field, and `cef-export.md`'s `devicePayloadId` mapping treats it the same way, but `audit-logs.md`'s Event schema table (which is supposed to be the canonical one) doesn't have it at all, not in the table, not in the example JSON, not in the hash-chain field list. Might be worth checking which one is actually right and fixing the other.

(careful: this comment is the result of some LLM analysis, treat it with caution)

documentation/docs/configuration/siems.md
```diff
@@ -94,7 +184,9 @@ field extraction, facets, or dashboards:
 | `object_uid` | string | Yes | Target object identifier (KMIP `UniqueIdentifier`) |
 | `algorithm` | string | Yes | Cryptographic algorithm (e.g. `AES`, `RSA`) |
 | `client_ip` | string | Yes | Source IP address |
```
HatemMn, line 186: `client_ip` here has no caveat attached to it, but on the audit-logs page it's documented as only trustworthy when `--audit-trusted-proxy-cidrs` is set up correctly, otherwise it's just the peer IP or an XFF value from a proxy nobody configured. This is exactly the table someone would use to wire up SOC dashboards and field mappings, so a short link back to that section would help people not treat this field as always validated.

(careful: this comment is the result of some LLM analysis, treat it with caution)

documentation/docs/configuration/siems.md
```diff
@@ -124,7 +216,38 @@ The `ckms audit export --format cef` command converts the JSONL audit store into
...
+```
+
+Send CEF events with octet-counting framing:
+
+```bash
+# Each frame: "<byte-count> <message>"
+# The message is a syslog PRI header + CEF line.
+while IFS= read -r line; do
+  msg="<134>$(date '+%b %d %H:%M:%S') kms-audit: ${line}"
+  printf '%zu %s' "${#msg}" "${msg}" > /dev/tcp/<host>/5514
```
HatemMn, line 241: This snippet won't actually run outside of bash. `printf '%zu %s' ...` throws `printf: %z: invalid directive` on zsh and on dash/sh, tested this live on both. The `< <(...)` process substitution a few lines below is already bash only too, so it might be worth adding a `#!/usr/bin/env bash` shebang or at least a note that this needs bash specifically, otherwise someone on macOS (zsh by default) or in a minimal container (dash) copy pastes this and nothing gets sent, with no clue why.

(careful: this comment is the result of some LLM analysis, treat it with caution)

.mise/scripts/test/rsyslog.conf
HatemMn, line 1: This file seems like a comment only stale artifact, delete ?

HatemMn (dismissed):
Overall, multiple updates introduced are unrelated to this and should be pushed to a new branch about "Monitoring refactoring"

The last comment, about the docker-file might be a real concern with the codebase that we should have eventually noticed, to be discussed in another future issue

documentation/docs/SUMMARY.md
HatemMn: deferred to https://github.com/Cosmian/kms/pull/1150

.github/workflows/test_all.yml
HatemMn: deferred to https://github.com/Cosmian/kms/pull/1150

documentation/nav.yml
HatemMn: deferred to https://github.com/Cosmian/kms/pull/1150

.mise/tasks/test/monitoring
HatemMn: deferred to https://github.com/Cosmian/kms/pull/1150

chokoblitz (commented):
Small heads-up, unrelated to the PR I've been working on: while writing a
similar live-fixture generator for the OpenSearch/Splunk compat checks, I
modeled it on this script and found a bug worth flagging here.

`ckms_run()` does `2>/dev/null || true`, which swallows every failure. Combined
with two CLI mismatches — `--output` isn't a valid flag (it's `--output-file`),
and `sym keys destroy` on an Active key is rejected by KMIP
(`Wrong_Key_Lifecycle_State`) unless it's `revoke`d first — the script prints
"Encrypted." / "Decrypted." / "Destroyed key." unconditionally and still
reports PASSED, but the actual audit file only ever gets 4 events: the `kmip`
422 startup probe, `Create`, `GetAttributes`, and a `Destroy` that itself fails
with 422. Encrypt, Decrypt, and the intended "export on non-existent key"
Failure never actually run. The guards only check total event count / hash
chain / required-fields presence, so none of them catch it.

Ran it standalone today to confirm it's still reproducible. Happy to send a
small patch if useful, otherwise just flagging for whenever it's convenient.

documentation/docs/configuration/otlp-telemetry.md
Manuthor: The file `otlp-telemetry.md` was already deleted and replaced by `otlp-metrics.md` in commit `abe55e6a5` when observability/monitoring content unrelated to audit was split off to `feat/other_changes_than_audit`. No references to `otlp-telemetry.md` remain in the repository.

.mise/tasks/test/audit
Manuthor: You are completely right. Audit logging (the `AuditMiddleware`, SHA-256 hash chaining, and JSONL writer) is available in both FIPS and non-FIPS builds and has no non-FIPS gating. The restriction and comment were erroneously copied over from the CEF test task boilerplate. The artificial restriction has been removed, defaulting the variant to `fips` while supporting both `fips` and `non-fips`.

lychee.toml
Manuthor: `lychee.toml` configures `lychee`, the Markdown link checker executed in CI (`linkcheck` job in `.github/workflows/main_base.yml`) against all documentation files under `documentation/docs/`.

This PR introduces extensive documentation for audit logging and SIEM integrations (`audit-logs.md`, `siems.md`, `cef-export.md`, and associated ADRs). Updating `lychee.toml` is necessary to prevent false-positive CI failures:
1. Placeholder/example domains like `kms.different.com` must be excluded so `lychee` does not attempt HTTP requests to nonexistent example domains.
2. External sites like `percona.com` throttle or block automated crawlers.

Therefore, modifying `lychee.toml` is directly tied to the documentation updates in this PR to ensure CI link checks pass.

documentation/nav.yml
Manuthor: The observability/OTel/monitoring documentation changes have already been split off into a separate branch (`feat/other_changes_than_audit`) in commit `abe55e6a5`. The remaining navigation entries in this PR only cover the Audit & SIEM documentation (`configuration/audit-logs.md`, `configuration/cef-export.md`, `configuration/siems.md`, and `kms_clients/audit.md`).

.mise/scripts/test/test_siem_cef_tcp_syslog.sh
Manuthor: Updated the framing calculation to compute the exact message byte length using `LC_ALL=C printf '%s' "${syslog_msg}" | wc -c | tr -d ' '` to adhere to RFC 6587 §3.4.1 octet-counting semantics across arbitrary multi-byte UTF-8 sequences, and replaced `%zu` with standard `%d` in `printf`.

.mise/scripts/test/test_siem_cef_tcp_syslog.sh
Manuthor: Updated `.mise/scripts/test/test_siem_cef_tcp_syslog.sh` to install `requirements-cef.txt` (`jc`) into the test virtualenv and invoke `cef_helper.py --jsonl "${AUDIT_JSONL}" --cef "${VERIFY_CEF}"`, replacing the ad-hoc bash verification with the full CEF v27 and `jc` round-trip validation.

documentation/docs/configuration/siems.md
Manuthor: Updated the Model 2 sequence diagram to include an explicit `Transport (shell / nc / logger)` actor. The diagram now clearly depicts `ckms` reading `audit.jsonl` and outputting CEF lines to stdout, while the separate shell pipeline/forwarder frames and delivers bytes over the wire (TCP RFC 6587 octet-counting or UDP) to the syslog/SIEM listener.

.github/workflows/test_all.yml
Manuthor: Already addressed in commit abe55e6a5 — the monitoring test entry was removed from the test matrix and deferred to the dedicated OTel/observability PR.

.github/workflows/test_all.yml
Manuthor: Good catch. Audit logging (AuditMiddleware, JSONL log writer, SHA-256 hash chaining) is fully FIPS-compatible and not gated behind `non-fips`. We have removed the matrix exclusion in `.github/workflows/test_all.yml`, so audit integration tests now run on both FIPS and non-FIPS builds.

documentation/docs/configuration/siems.md
Manuthor: Updated the rsyslog example configuration to use `template="RSYSLOG_FileFormat"`. `RSYSLOG_FileFormat` appends `\n` after each message, ensuring one line per event for omfile output, aligning with the ADR and preventing concatenated lines.

documentation/docs/configuration/audit-logs.md
Manuthor: The sequence diagram referencing internal middleware implementation details has already been removed from `documentation/docs/configuration/audit-logs.md`.

documentation/docs/configuration/audit-logs.md
Manuthor: Already addressed. The retention guidance has been moved to the **Best practices** section, introducing the general retention practice first and citing PCI-DSS Req. 10.7 (12 months) and HIPAA §164.312(b) as examples. In addition, all SIEM integration details have been moved to `siems.md` with complete end-to-end configs, leaving only a clean cross-reference link here.

documentation/docs/configuration/audit-logs.md
Manuthor: Agreed. The redundant in-line SIEM section was removed from `audit-logs.md`. All SIEM ingestion patterns and configurations are consolidated in `siems.md`, with only a concise reference link retained in `audit-logs.md`.

documentation/docs/configuration/audit-logs.md
Manuthor: Complete end-to-end configurations for Splunk (inputs.conf), Elasticsearch/Filebeat (filebeat.yml), Fluent Bit (fluent-bit.conf), and rsyslog (imfile and imtcp) are now provided directly in `documentation/docs/configuration/siems.md`.

.mise/scripts/test/otlp-audit-collector-config.yaml
Manuthor: The file was already removed in commit `abe55e6a5` when observability-related test artifacts were split off.

.mise/scripts/test/test_audit_log.sh
Manuthor: Added a tamper-evidence test (GUARD 2b) to `.mise/scripts/test/test_audit_log.sh`. It copies the generated audit log to a temporary file, mutates an event's `operation` field on line 2 (invalidating its SHA-256 row hash), and asserts that `ckms audit verify --path` detects the tampering and fails with a non-zero exit code. The temporary tampered file is cleaned up after verification and registered in the script's exit trap.

documentation/docs/SUMMARY.md
Manuthor: All unrelated OTel edits have already been split off to `feat/other_changes_than_audit` in commit `abe55e6a5`. Only Audit & SIEM docs remain in `SUMMARY.md`.

.mise/tasks/test/siem
Manuthor: Updated `.mise/tasks/test/siem` to automatically start Elasticsearch via `docker compose up -d elasticsearch` before executing the `all` or `filebeat` test suites if not already running (or fail fast if Docker is unavailable). Also updated `.mise/scripts/test/test_siem_filebeat.sh` to check if Elasticsearch is reachable or the container is running via `docker compose ps --status running -q elasticsearch`, failing fast immediately with setup instructions instead of waiting for the 120s timeout.

documentation/docs/adr/2026-07-09-audit-log-single-writer-design.md
Manuthor: Already addressed in commit `5b9de3c41` ("chore: drop ADR references from test scripts (ADR filenames/numbers churn too much to hardcode)"), which removed all stale numeric ADR references (such as ADR-003, ADR-004, ADR-005, and ADR-007) from `.mise/tasks/test/` and test scripts.

docker-compose.yml
Manuthor: Added `profiles: [test, siem]` to the `elasticsearch` service so it will never start during a bare `docker compose up`, requiring explicit startup (`docker compose up -d elasticsearch`) or `--profile test`/`--profile siem`. Added a prominent security warning stating that this service is strictly for local tests and CI automation with authentication and TLS disabled.

.mise/tasks/test/siem-filebeat
Manuthor: Updated `test_siem_filebeat.sh` error handling when indexing times out: generalized the diagnostic message to cover both pipeline/mapping errors and indexing blocks (such as disk flood-stage watermarks), queried `_cluster/health` (including status, unassigned shards, and allocation explain), and printed the last 25 lines of Filebeat container logs before container removal.

.mise/tasks/test/monitoring
Manuthor: Already addressed in commit `abe55e6a5`: the monitoring task changes were split off to `feat/other_changes_than_audit`.

.mise/scripts/README.md
Manuthor: The previous document was legacy documentation detailing the obsolete `bash nix.sh <command>` workflow. Project automation (building, testing, packaging, CI) has transitioned to MISE tasks (`.mise/tasks/`). The file was updated to provide an accurate, maintainable reference for the active task infrastructure, documenting `mise run <task>` targets (`build:*`, `test:*`, `release:*`, `package:*`) and their flags (`--variant`, `--link`, etc.) as used today.

documentation/docs/configuration/siems.md
Manuthor: Verified against `crate/access/src/audit/event.rs` and `hash.rs`: `request_id` is indeed a real field on `AuditEvent` used for correlation across operations in the same request and included in the canonical byte hashing. Updated `documentation/docs/configuration/audit-logs.md` to document `request_id` in the Event schema table, add it to the example JSON event, and reflect it in the canonical byte sequence in the hash chain description.

documentation/docs/configuration/siems.md
Manuthor: Updated the `client_ip` description in the SIEM field reference table to clarify that it represents the direct TCP peer unless configured behind trusted proxies, and added a link to the `Client IP and reverse proxies` section in `audit-logs.md`.

documentation/docs/configuration/siems.md
Manuthor: Updated the TCP syslog snippet with a `#!/usr/bin/env bash` shebang and notes explaining the Bash requirement (`/dev/tcp` redirection and `< <(...)` process substitution). Replaced the non-standard `%zu` directive with standard `%d`, computed the exact byte count via `LC_ALL=C printf "%s" "${msg}" | wc -c | tr -d " "`, and noted `nc <host> 5514` as an alternative to `/dev/tcp`.

.mise/scripts/test/rsyslog.conf
Manuthor: This file is not comment-only or a stale artifact. While lines 1–9 provide explanatory header comments, lines 11–20 define the active rsyslog configuration: loading the `imtcp` module, binding the TCP port 514 listener, and routing messages to `/var/log/kms-cef.log` via the `kms_cef` ruleset. It is actively mounted into the test container in `.mise/scripts/test/test_siem_cef_tcp_syslog.sh` (`-v "${SCRIPT_DIR}/rsyslog.conf:/etc/rsyslog.conf:ro"`) to validate RFC 6587 octet-counting CEF delivery over TCP syslog. Deleting it would break the integration test suite.

Manuthor:
Thanks @chokoblitz! Excellent catch.

Fixed across all 7 SIEM/audit integration test scripts (`test_audit_log.sh`, `test_cef.sh`, `test_cef_format.sh`, `test_siem_cef_syslog.sh`, `test_siem_cef_tcp_syslog.sh`, `test_siem_filebeat.sh`, and `test_siem_fluent_bit.sh`):

1. **CLI flag correction**: Replaced `--output` with `--output-file` in `sym encrypt` and `sym decrypt`.
2. **Lifecycle state compliance**: Added `sym keys revoke "test cleanup"` before `sym keys destroy` so symmetric keys transition from Active -> Revoked -> Destroyed, succeeding properly.
3. **Deliberate failure export**: Provided `<KEY_FILE>` positional argument (`/dev/null`) to `sym keys export --key-id "00000000-0000-0000-0000-000000000000"` so clap parsing succeeds and reaches the KMS server to produce the intended 422 Failure audit event.
4. **Error propagation**: Removed `2>/dev/null || true` from `ckms_run()` so any unexpected non-zero exit code immediately aborts the test script under `set -euo pipefail`. Added a dedicated `ckms_run_fail()` helper to assert expected failure on negative tests.
5. **Python 3.9 compatibility**: Normalized fractional ISO timestamps in `cef_helper.py` so microsecond parsing works across Python 3.9–3.12+.

All tests verified live against the server (`test:audit`, `test_cef_format.sh`, `test_cef.sh` with 14/14 events validated against `jc`).

HatemMn:
@Manuthor just for info `feat/audit_startup `just got rebased onto current develop

PR #1121, opened 2026-08-12
feat: add pgsql for auditing

### LINKED ISSUE #1055: Phase 1: PART C: Audit databases (PostgreSQL / Sqlite )
> [audit.db] — optional. SQLite or PostgreSQL, separate from the key DB. KMS connects with a write-only credential (INSERT only). ckms audit export uses a separate read-only credential.

Due to the code being already complex enough, this task has been differrent

Both and PostgreSQL has both been confirmed to be good choices

**PostGres** being a defacto standard and a very advanced DB that operators typically use for logging, it will be the top priority

HatemMn:


HatemMn:
https://github.com/Cosmian/kms/pull/1186

PR #1163, opened 2026-09-02
fix(deps): bump brace-expansion and nanoid to patch UI CVEs

Svector-anu:
Automated dependency bump to address disclosed CVEs in the UI lockfile.

- **CVE:** CVE-2026-14257 / CVE-2026-69152
- **Advisory:** https://github.com/advisories/GHSA-mh99-v99m-4gvg · https://github.com/advisories/GHSA-rgw5-rvv9-x895
- **Severity:** high
- **Package:** `brace-expansion` `5.0.7` → `>=5.0.9` (pnpm override)

- **CVE:** CVE-2026-67213
- **Advisory:** https://github.com/advisories/GHSA-2v37-7h3g-55p8
- **Severity:** high
- **Package:** `nanoid` `3.3.16` → `3.3.18` (pnpm override, pinned to the 3.x patched line)

Detected by [osv-scanner](https://google.github.io/osv-scanner/). No code changes outside `ui/package.json` overrides and `ui/pnpm-lock.yaml`.

### Verification
- Reproduced locally: yes
- Command: `osv-scanner scan source --recursive --no-ignore --format=json .` then inspected `ui/pnpm-lock.yaml` for `brace-expansion@5.0.7` / `nanoid@3.3.16`
- Before: lockfile resolved `brace-expansion@5.0.7` and `nanoid@3.3.16` (override floor was `>=5.0.7`)
- After: overrides require `brace-expansion>=5.0.9` and `nanoid` `3.3.18`; lockfile refreshed with `pnpm install --no-frozen-lockfile`
- Environment: osv-scanner 2.5.1, pnpm 10.17.1

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -73,7 +73,8 @@
         "overrides": {
             "esbuild": "^0.28.1",
             "ws": ">=8.20.1",
-            "brace-expansion": ">=5.0.7",
+            "brace-expansion": ">=5.0.9",
+            "nanoid": "3.3.18",
             "@babel/runtime": "^7.26.10",
             "minimatch": ">=10.2.1",
             "picomatch": ">=4.0.4",
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -7,7 +7,8 @@ settings:
 overrides:
   esbuild: ^0.28.1
   ws: '>=8.20.1'
-  brace-expansion: '>=5.0.7'
+  brace-expansion: '>=5.0.9'
+  nanoid: 3.3.18
   '@babel/runtime': ^7.26.10
   minimatch: '>=10.2.1'
   picomatch: '>=4.0.4'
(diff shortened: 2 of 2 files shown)


Manuthor:
Thanks for preparing these dependency updates. A few suggestions before merging:

- Please frame this as dependency/build-tool hardening rather than a high-severity KMS runtime security fix: `brace-expansion` is reached through ESLint tooling, while `nanoid` is reached through PostCSS/Vite, with no identified untrusted KMS input reaching the vulnerable functions.
- Please regenerate `ui/pnpm-lock.yaml` with the repository-pinned `pnpm@10.17.1` to avoid the unrelated removal of `libc` metadata.
- Since the lockfile changes, please update the Nix pnpm hashes. Run `nix-build -A ui-fips` on both Linux and macOS, copy each reported `got: sha256-...` value into `nix/expected-hashes/ui.pnpm.linux.sha256` and `nix/expected-hashes/ui.pnpm.darwin.sha256`, then rerun the command to verify it succeeds.
- Please also verify `pnpm install --frozen-lockfile` and the relevant UI checks.

The version bumps themselves look appropriate. Thank you!

ui/package.json
```diff
@@ -73,7 +73,8 @@
         "overrides": {
             "esbuild": "^0.28.1",
             "ws": ">=8.20.1",
-            "brace-expansion": ">=5.0.7",
+            "brace-expansion": ">=5.0.9",
+            "nanoid": "3.3.18",
```
HatemMn, line 77: @Svector-anu can you please briefly explain why a new depency needed to be added for this patch despite not seeing it in other diffs ?

Svector-anu:
Good question. `nanoid` isn't a direct SDK dependency — it's pulled in transitively via `postcss` (resolving to the vulnerable 3.3.16). Since it's transitive, `overrides` is the standard pnpm mechanism to force the whole tree onto a patched version without touching `package.json`'s real dependency list — same pattern as the existing `brace-expansion` override in this same block, just for a different CVE (`CVE-2026-67213` / GHSA-2v37-7h3g-55p8, HIGH).

Separately, worth being upfront about something I checked after seeing a reviewer catch the same class of issue on a sibling PR: this update also drops `libc` platform constraints on ~41 unrelated packages in the lockfile. I dug into it — it's not scope creep in how the bump was done. I reproduced it with the exact pinned `pnpm@10.17.1` (not a version mismatch), confirmed a completely untouched `pnpm install --frozen-lockfile` is a true no-op, and confirmed it's specifically re-resolving these two packages that triggers it — pnpm's fresh resolution against the current registry just doesn't reproduce those legacy `libc` annotations for those particular entries anymore. I also tried hand-patching them back in and caught myself corrupting a couple of libc values in the process (glibc/musl swapped on adjacent entries) — reverted that, since a wrong libc tag is worse than a missing one. So this lockfile is genuinely what the pinned tooling produces today; not something I can tighten further without inventing data pnpm itself won't stand behind.

PR #1164, opened 2026-09-03
feat: add OCSP responder support

Manuthor:
### Add OCSP responder support and Validate lifecycle-state checks
- Adds public GET (`/ocsp/{encoded}`) and POST (`/ocsp/`) OCSP endpoints, an OpenSSL-backed response builder, and a process-wide response cache configured via `OcspConfig`.
- `Validate` operation now tracks internal KMS certificate lifecycle state and rejects chains containing certificates that are not Active or PreActive, with a 32-entry request limit.
- Certificate revocation now evicts the corresponding OCSP cache entry to prevent stale responses.
- Adds black-box OCSP and PKI revocation test tasks in `.mise/tasks/test/` and updates documentation.
- Behavioral Change: `Validate` in [validate.rs](https://github.com/Cosmian/kms/pull/1164/files#diff-2697f02be55d72fcd1ae037c2cb0fe39766c6ecf75577ed99ad4470aa5493591) rejects tracked certificates in Compromised or Deactivated states. Network-dependent tests were removed from [test_validate.rs](https://github.com/Cosmian/kms/pull/1164/files#diff-4ab1c87a6f2d792d18e42d81f7ad868979c3ecbcd1bcf35f85a30f2b88c0e938) and replaced with internal-state and CRL cascade tests.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized f0d3415.</sup>


serene-kitfisto-8899:
## Security review: OCSP responder implementation

I reviewed the OCSP responder code (`crate/crypto/src/openssl/ocsp.rs`, `crate/server/src/routes/ocsp/handler.rs`, `crate/server/src/config/command_line/ocsp_config.rs`, and the route registration in `start_kms_server.rs`) against a checklist of known OCSP protocol/implementation security issues (replay, nonce handling, soft-fail/stale responses, signing algorithm choice, and DoS).

### ✅ Handled well

| Checklist item | Evidence |
|---|---|
| Nonce validated/echoed, no memory blow-up | `extract_request_nonce`/`handle_nonce` enforce the RFC 9654 16–128 octet bound; out-of-range nonces are silently ignored rather than stored/echoed; `Required` policy cleanly rejects missing nonces via `malformedRequest` (`crate/crypto/src/openssl/ocsp.rs:607-730`). |
| FIPS-approved signing algorithm | `signing_md()` uses SHA-256 for RSA/EC and null-digest for PQC (self-hashing algorithms) — `crate/crypto/src/openssl/ocsp.rs:946`. |
| `thisUpdate`/`nextUpdate` freshness | Both always set from `now`/`now+ttl_secs` (`ocsp.rs:373-395`); the in-memory cache independently re-checks `expires.elapsed().as_secs() < ttl_secs` before serving (`handler.rs:250-251`), so cached entries can't outlive their own signed validity window. |
| Malformed-request/DoS hardening | Query count bounded (`MAX_OCSP_QUERIES_PER_REQUEST = 128`, `ocsp.rs:49,244-249`) and rejected *before* allocation/DB/signing work; malformed DER returns a proper OCSP-wire `malformedRequest` response instead of crashing; routes inherit the app's global `RateLimiterMiddleware` and 64MB `PayloadConfig` (`start_kms_server.rs:1148-1151,1174`). |
| No stale/replayed "good" responses across revocation | `evict_ocsp_cache_entry` is called immediately from `revoke.rs` on state change (`handler.rs:78-89`); CA-compromise cascade unconditionally bypasses the cache for all entries once the CA itself is compromised (`handler.rs:237-241`). |

### ⚠️ Gaps found

1. **No `extKeyUsage: OCSPSigning` / `id-pkix-ocsp-nocheck` validation on the delegated responder certificate** (RFC 6960 §4.2.2.2).
   The doc comment on `ocsp_responder_cert_uid` (`ocsp_config.rs:43-51`) *documents* that the referenced cert "MUST have" the `OCSPSigning` EKU and the `nocheck` extension, but `retrieve_signer_cert_and_key()` (`handler.rs:389-489`) never actually checks these on the certificate object — it only checks the KMS lifecycle `State` (Compromised/Destroyed_Compromised). An operator could point `ocsp_responder_cert_uid` at any cert/key pair in the KMS (e.g. a TLS server cert) and the responder would sign with it anyway. Suggest adding an explicit EKU/extension check to enforce the documented invariant and match RFC 6960 compliance.

2. **No explicit size cap on the GET path before base64/DER decoding.**
   RFC 6960 Appendix A recommends the GET method only for requests ≤255 bytes. `get_ocsp()` (`handler.rs:107-119`) base64url-decodes the path segment with no OCSP-specific upper bound before doing so — only Actix's generic URI length limit applies. Low severity given the downstream `MAX_OCSP_QUERIES_PER_REQUEST` bound, but a cheap explicit length guard (e.g. reject > ~4KB before decoding) would close this off entirely and align with the RFC's intent.

3. **No OCSP-specific rate limiting beyond the general per-IP limiter.**
   The endpoints are correctly public/unauthenticated (RFC 6960 §2), and they do inherit the global `RateLimiterMiddleware`, but it's keyed by peer IP only. A distributed low-and-slow query flood (many unique serials, or many source IPs) can still force repeated signing/DB work, since the cache is single-entry-per-serial and is bypassed whenever a nonce is present or the CA is compromised. Not a regression, just a residual capacity risk worth noting given OCSP signing (especially HSM-backed) is comparatively expensive, per the module's own "minimise signing key usage" design goal.

None of these are remotely exploitable critical bugs — the implementation already correctly handles the classic OCSP pitfalls (nonce bounds/replay, cache invalidation on revoke, CA-compromise cascade, stale-response avoidance, request-size DoS via query count). Gap #1 (delegated-responder EKU enforcement) is the one I'd recommend prioritizing since the code's own comments assert an invariant that isn't currently enforced.

serene-kitfisto-8899 (approved):
Good to go.

PR #1189, opened 2026-09-11
[Audit Sink][PART1] Introduce Audit Sink Trait

HatemMn:
_reminder of the order_

ci/audit-jsonl-compat
  └─ feat/audit-sink-refactor        (4 commits)
       └─ feat/audit-sink-generalize (1 commit)
            └─ feat/audit-postgres-backend (2 commits)
                 └─ feat/audit-postgres-cli (3 commits)

---

# Overview

**This PR is just a refactoring/a split** for the 2090-line audit file store into three modules before touching anything real:

`file_sink.rs` (main file), writer.rs (the writer loop), store.rs (the handle + tests), `recovery.rs` (recovery).

Also pulled event construction into one place in access, split the fat `middlewares/audit.rs` into `mod.rs` + `client_ip.rs` + `extensions.rs`, and added the new `AuditSink` trait in interfaces (unused so far).

No behavior change anywhere in this PR. 

**You might notice that two audit traits are living separatly: one of them is the legacy file trait, it will be removed in the next PR**


crate/server/src/core/audit/file_sink.rs
```diff
@@ -0,0 +1,949 @@
...
+
+    Ok((next_id, prev_hash))
+}
+
+/// Abstraction over the audit log's underlying writer.
+///
+/// This exists so the fault path (`write_event` failing mid-run) can be
+/// exercised in tests with a mock sink, without touching real files —
+/// production always uses `std::fs::File`.
+pub(super) trait AuditSink {
```
HatemMn, line 64: Just saying: this is here only because the next PR will follow up and deletes it, refactoring with the main "new trait"

crate/interfaces/src/stores/audit_sink.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use async_trait::async_trait;
+use cosmian_kms_access::audit::AuditEvent;
+
+use crate::InterfaceResult;
+
+/// Position of the audit hash chain: the id to assign to the next event, and the
+/// `row_hash` of the last durably persisted one.
+#[derive(Debug, Clone, Copy)]
+pub struct ChainHead {
```
Manuthor, line 26:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
pub struct ChainHead {
```
Derive `PartialEq`, `Eq`, and `Default` on `ChainHead` so instances can be directly compared (e.g. `head == ChainHead::EMPTY`) and defaulted in consumers and unit tests.

crate/interfaces/src/stores/audit_sink.rs
```diff
@@ -0,0 +1,77 @@
+//! The `AuditSink` trait: a durable destination for finalised audit events.
+//!
+//! Implemented by each backend that wants to persist the audit hash chain. A sink never
+//! assigns ids and never computes hashes — it persists what it is given, in the order it
+//! is given, and reports where the chain left off so the writer can resume it. Backends
+//! are interchangeable at the trait boundary: a chain started on one backend can be
+//! verified after export from another, because both encode the same [`AuditEvent`] and
+//! the same canonical hash (see `cosmian_kms_access::audit::canonical_bytes`).
```
Manuthor, line 8:
```suggestion
//! the same canonical hash (see `cosmian_kms_access::audit::compute_row_hash`).
```
`canonical_bytes` is private (`pub(crate)`) inside `cosmian_kms_access::audit::hash` and not re-exported. Public callers and documentation should reference `compute_row_hash` or `cosmian_kms_access::audit`.

crate/server/src/core/audit/recovery.rs
```diff
@@ -0,0 +1,733 @@
...
+    let Some(anchor_line) = second_last else {
+        // Can't happen per the invariant above, but never panic on a recovery path.
+        return Ok(TailOutcome::SealAndRoll {
+            reason: SealReason::Unparsable,
+            claimed_last_id: None,
+            failure_offset: last_start,
+        });
+    };
+    match check_row(anchor_line) {
+        RowCheck::Verified(event) => match event.id.checked_add(1) {
```
Manuthor, line 321:
```suggestion
    match check_row(anchor_line) {
        RowCheck::Verified(event) => {
            if !verify_chain_link(&event, previous_event) {
                return Ok(TailOutcome::SealAndRoll {
                    reason: SealReason::ChainBroken,
                    claimed_last_id: Some(event.id),
                    failure_offset: last_start,
                });
            }
            match event.id.checked_add(1) {
                Some(next_id) => Ok(TailOutcome::TruncateContinue {
                    keep_len: last_start,
                    next_id,
                    prev_hash: event.row_hash,
                    bytes_discarded: last_end - last_start,
                    discard_offset: last_start,
                }),
                None => Ok(TailOutcome::SealAndRoll {
                    reason: SealReason::IdOverflow,
                    claimed_last_id: Some(event.id),
                    failure_offset: last_start,
                }),
            }
        }
        RowCheck::HashMismatch(_) | RowCheck::Unparsable => Ok(TailOutcome::SealAndRoll {
```
When recovering from a torn trailing write, verify `verify_chain_link(&event, previous_event)` on the fallback anchor event before accepting it as the truncation boundary. If the fallback anchor row itself has a broken hash link, `classify_tail` should trigger `SealAndRoll` rather than resuming over a corrupted chain.

crate/interfaces/src/stores/audit_sink.rs
HatemMn: Okay with the pub part, but reading back `compute_row_hash` : 

```rust
/// Computes the SHA-256 digest over the canonical byte form of `event`.
/// The returned hash should be stored as `event.row_hash`.
#[must_use]
pub fn compute_row_hash(event: &AuditEvent) -> [u8; 32] {
    let mut hasher = Sha256::new();
    hasher.update(canonical_bytes(event));
    hasher.finalize().into()
}
```

is just a wrapper. The dev who made the effort to look for it will just go jump into `cannonical_bytes`, that comment spares him this intermediary step.

This comment **does hold** for libraries whose docs are published on crates.io - in my knowledge this crate is no such candidate, so anyone reading the above comment has probably pulled all the codebase anyway.

If `audit_sink.rs` will appear on crates.io please tell me, otherwise we can close this comment

crate/interfaces/src/stores/audit_sink.rs
HatemMn: Gladly adding Eq and PartialEq

Not adding Default for now (we have chainHead::EMPTY already, and I prefer avoiding the "two ways to do the same thing" in the code)

crate/server/src/core/audit/recovery.rs
HatemMn: If we apply this suggestion, I fear that the following regression will happen: every ordinary torn-write recovery (a ordinary crash mid-fsync, the exact case this function exists to self-heal) would now hit SealAndRoll(ChainBroken) instead of TruncateContinue, because the new check can never pass. That silently disables the "always start, self-heal from a crash" guarantee documented at the top of the file, and would generate false tampering alerts (plus needless file-sealing/renaming) on every restart-after-crash.

I suggest to keep as-is, unless we need to discuss changing the contract of `TruncateContinue`

```rust
    /// A torn (interrupted) write: the last row is incomplete, but the row before it
    /// (or genesis) is trustworthy. Truncate the torn fragment and continue in place.
    TruncateContinue {
        /// Byte length to truncate the file to — always the end of the last verified row.
        keep_len: u64,
        next_id: i64,
        prev_hash: [u8; 32],
        bytes_discarded: u64,
        discard_offset: u64,
    },
```

Servers (ie. the KMS) are very usually restarted by random agents or humans, torn audit files will for sure rise during the lifecycle of a KMS, and defaulting to SealAndRoll routine while _we can_ avoid doing it avoids the admin to have to handle tens of sealed old artifacts

HatemMn:
image

crate/interfaces/src/stores/audit_sink.rs
Manuthor: Fair point on discoverability for someone who already has the whole repo checked out — but that's not quite the objection. canonical_bytes is pub(crate) inside cosmian_kms_access::audit::hash and isn't re-exported (see the pub use list in crate/access/src/audit/mod.rs). So cosmian_kms_access::audit::canonical_bytes isn't a path any other crate — interfaces or server included — can actually name; it wouldn't compile as an import. 

This isn't "the reader takes one extra jump," the doc points at a symbol that's structurally unreachable from where the comment lives, independent of [crates.io](http://crates.io/). Let's repoint to compute_row_hash (the original suggestion), or re-export canonical_bytes explicitly if we want the lower-level fn named in the doc.

crate/interfaces/src/stores/audit_sink.rs
Manuthor: No objection to keeping EMPTY as the canonical constructor — but "two ways to do the same thing" only applies if Default could diverge from it. 

`impl Default for ChainHead { fn default() -> Self { Self::EMPTY } }` is a pure delegate, zero duplication risk, and it buys interop with generic code that wants `Default` specifically (for example `Option<ChainHead>::unwrap_or_default()` or any future `#[derive(Default)]` struct embedding ChainHead) without callers needing to know EMPTY exists.

crate/server/src/core/audit/recovery.rs
Manuthor: Correct. Withdrawn

HatemMn:
For info: this **https://github.com/Cosmian/kms/pull/1190**

Got merged here for less PR overhead

PR #1190, opened 2026-09-11
[AUDIT SINK PART 2] File implements audit Sink Trait

HatemMn:
# Overview

**`feat/audit-sink-generalize`**

One commit. This is where the file backend actually adapts to the new public `AuditSink` trait instead of a private mock one. 

Small diff but this is the one place old behavior could have silently shifted during the refactor.

Got its own PR instead of getting buried in the split commit before it.


crate/server/src/core/audit/file_sink.rs
```diff
@@ -56,146 +38,229 @@ impl AuditWriteState {
...
+    file: &mut std::fs::File,
+    draft: AuditEventDraft,
+    next_id: i64,
+    prev_hash: &mut [u8; 32],
+) -> i64 {
+    let event = draft.finalize(next_id, *prev_hash);
+    match write_event_line(file, &event) {
+        Ok(()) => {
+            *prev_hash = event.row_hash;
+            next_id.checked_add(1).unwrap_or(next_id)
```
Manuthor, line 61: In `write_recovery_sentinel()`, if `next_id` reaches `i64::MAX`, `next_id.checked_add(1).unwrap_or(next_id)` silently leaves `next_id` at `i64::MAX` without logging an overflow error (unlike `write_draft_to_chain()` in `writer.rs` which explicitly logs an `error!` when `next_id` overflows).

```suggestion
            next_id.checked_add(1).unwrap_or_else(|| {
                error!("AuditFileStore: recovery sentinel id counter overflow at i64::MAX");
                next_id
            })
```

crate/server/src/core/audit/store.rs
```diff
@@ -97,7 +72,25 @@ impl AuditFileStore {
...
+                    writer_loop(
+                        sink,
+                        chain_head.next_id,
+                        chain_head.prev_hash,
+                        rx,
+                        dropped_count_for_writer,
+                    )
+                    .await;
+                }
+                Err(e) => {
```
Manuthor, line 87: In `AuditFileStore::start_with_max_size()`, if `sink.resume().await` returns an error (e.g. if an unforeseen non-recoverable error occurs in a third-party sink), the error is logged and the spawned task terminates. However, any callers who subsequently call `enqueue()` will have their events queued up to channel capacity and then dropped with channel full warnings. Consider exposing or tracking a sink failure state so `enqueue()` or health checks can distinguish between a busy channel and a dead writer task.

Manuthor (commented):
Code review for PR #1190 (`feat/audit-sink-generalize` against `feat/audit-sink-refactor`):

The generalization of `AuditStore` over the async `AuditSink` trait is clean and sound. Moving recovery and exclusive lock handling into `FileSink::resume()` preserves the non-blocking startup contract. Below are specific in-scope suggestions.

crate/server/src/core/audit/store.rs
HatemMn: > Consider exposing or tracking a sink failure state so enqueue() or health checks can distinguish between a busy channel and a dead writer task.

Implementing that function wasn't hard but did you in an indirect manner suggest to modify `health.rs` by adding the audit as a dependency ? I see that database is already added as dependency

image

I will push my code to another branch, please tell me this is what your required or not because I am not so sure (without that, the 'addition' will become dead code)

HatemMn: Test

HatemMn: https://github.com/Cosmian/kms/commit/8d589bb43fbbcac566a6163ba99e7ac685b8c132

HatemMn (commented):
done

crate/server/src/core/audit/file_sink.rs
```diff
@@ -56,146 +38,232 @@ impl AuditWriteState {
...
+                Ok(meta) => break meta.len(),
+                Err(e) => {
+                    error!(
+                        "AuditFileStore: cannot stat recovered log file {} ({e}) — retrying",
+                        self.path.display()
+                    );
+                    tokio::time::sleep(OPEN_RETRY_INTERVAL).await;
+                }
+            }
+        };
```
Manuthor, line 197: If file.metadata() fails with a handle-scoped error such as ESTALE on a shared volume or network mount, retrying on the existing file handle loops indefinitely while holding the exclusive lock. This prevents writer_loop from starting, saturates the bounded channel, and eventually drops audit events.

Consider fetching metadata as part of the blocking recovery task so a stat error re-runs recovery and reopens the path rather than spinning on the same descriptor:

```suggestion
        let (file, next_id, prev_hash, committed_len) = loop {
            let path_for_recovery = self.path.clone();
            let recovered = tokio::task::spawn_blocking(move || {
                let (file, next_id, prev_hash) = recover_and_open(&path_for_recovery)?;
                let len = file.metadata()?.len();
                Ok::<_, std::io::Error>((file, next_id, prev_hash, len))
            })
            .await;
            match recovered {
                Ok(Ok(quad)) => break quad,
                Ok(Err(e)) => {
                    error!(
                        "AuditFileStore: cannot open or stat audit log {} ({e}) — retrying",
                        self.path.display()
                    );
                    tokio::time::sleep(OPEN_RETRY_INTERVAL).await;
                }
                Err(join_err) => {
                    error!("AuditFileStore: recovery task failed to run ({join_err}) — retrying");
                    tokio::time::sleep(OPEN_RETRY_INTERVAL).await;
                }
            }
        };

        self.committed_len = committed_len;
```

crate/server/src/core/audit/store.rs
Manuthor: Modifying health.rs wasn't required for this PR, but your implementation in that commit actually fits cleanly.                                                 
                                                                                                                                                                
 The primary goal of the comment was simply inside store.rs: when sink.resume().await fails and the task exits, callers keep calling enqueue(), filling the bounded channel until it starts dropping events with generic "channel full" logs. Tracking when the writer task is dead allows enqueue() to fail immediately or log accurately (e.g. audit writer task has exited).                                                                                                         
                                                                                                                                                                
Wiring this state into /health (as you did in 8d589bb) is a good addition that you can add directly to this PR.
When you will have fix this https://github.com/Cosmian/kms/pull/1190#discussion_r4070194961, you can merge the PR

crate/server/src/core/audit/file_sink.rs
HatemMn: thanks fixed

PR #1191, opened 2026-09-11
[AUDIT SINK PART 3] Feat/audit postgres backend

HatemMn:
# Overview

New Postgres audit backend plus the wiring to select it. `PgAuditSink` in `server_database`: schema with append-only triggers that reject `UPDATE`/`DELETE`/`TRUNCATE`, connection pool, retry logic. Wiring side: `--audit-postgres-url` / `--audit-instance-id` flags, `AuditBackendParams` enum, and `create_audit_store` in `kms/mod.rs` becomes async so it can connect, take a session advisory lock, and verify the whole existing chain before the server starts serving traffic. Any failure there aborts startup, doesn't leave the KMS running unaudited.

Config also blocks the Postgres audit URL from pointing at the same database as `--database-url`, and requires `--audit-instance-id` explicitly, no hostname guess, a wrong guess risks two instances sharing one chain.

Kept these two together on purpose. The schema and sink code are dead until the wiring commit lands, splitting them into two PRs means reviewing dead code in isolation for no reason. Highest logical stakes of the four, this is where a real bug would live.

## Additional info:

- **Sealing algorithm:** Each generation restarts at id = 0 rather than continuing the id sequence, because verify_chain_link accepts a genesis row only when current.id == 0 && current.prev_hash == [0u8; 32]. A re-anchor at id = N would fail verification under the shared verifier used by both backends and by ckms audit verify. Keeping id semantics identical across file and PostgreSQL is what makes one verifier correct for both — which in turn is why chain_generation must enter the primary key, since (instance_id, id) alone cannot hold two rows with id = 0.


HatemMn:
image

